Issue #28, opened 2022-01-20
Junk between ID3 data and MP3 frames is not skipped


localthomas:
I am unsure whether this is a bug or a feature request.
When junk data is between an ID3 data block and the following MP3 data, the `Probe` does not report the file type `MP3`.

Most other software I tested with simply skips the junk bytes (e.g. VLC, FFmpeg, MPV) without issues. Note that junk, in this case, does not mean the optional padding after ID3 data. The issue is with a "wrong" length field of the ID3 header.

Example code that panics on `guess_file_type()`:
```rust
// test data that contains 4 bytes of junk (0x20) between the ID3 portion and the first MP3 frame
let data: [&[u8]; 4] = [
	// ID3v2.3 header (10 bytes)
	&[0x49, 0x44, 0x33, 0x03, 0x00, 0x00, 0x00, 0x00, 0x00, 0x23],
	// TALB frame
	&[
		0x54, 0x41, 0x4C, 0x42, 0x00, 0x00, 0x00, 0x19, 0x00, 0x00, 0x01, 0xFF, 0xFE, 0x61,
		0x00, 0x61, 0x00, 0x61, 0x00, 0x61, 0x00, 0x61, 0x00, 0x61, 0x00, 0x61, 0x00, 0x61,
		0x00, 0x61, 0x00, 0x61, 0x00, 0x61, 0x00,
	],
	// 4 bytes of junk
	&[0x20, 0x20, 0x20, 0x20],
	// start of MP3 frame (not all bytes are shown in this slice)
	&[
		0xFF, 0xFB, 0x50, 0xC4, 0x00, 0x03, 0xC0, 0x00, 0x01, 0xA4, 0x00, 0x00, 0x00, 0x20,
		0x00, 0x00, 0x34, 0x80, 0x00, 0x00, 0x04,
	],
];
let data: Vec<u8> = data.into_iter().flatten().cloned().collect();
let data = std::io::Cursor::new(&data);
let probe = Probe::new(data);
let probe = probe.guess_file_type().unwrap();
matches!(probe.file_type(), Some(crate::FileType::MP3));
```

The issue is that only a simple 'if'-match is implemented for probing the sync bits of an MP3 frame:

https://github.com/Serial-ATA/lofty-rs/blob/c27264f50b0a2cc6285960c368a14457d9b0432c/src/probe.rs#L180-L186

[...]

PR #29, opened 2022-01-21
Byte skipping between ID3 and MP3 blocks

localthomas:
As discussed in #28, junk bytes between initial metadata blocks and following MP3 frames are skipped with this implementation change.

The new function `search_for_frame_sync` uses a reader to search for the first appearence of frame sync bits. This function is then used for probing the file type (`Probe::guess_inner`) and reading MP3 data (`mp3::read::read_from`). All three additions are checked by three new tests. The tests for the new function and the `Probe` are in the same module as the implementation, while the `mp3::read_from` test is separted into the `tests` and uses an example `*.mp3` file.

Please comment on the positioning of the tests and the changes on the two existing functions.

---

I am a bit confused by the `Probe::guess_inner` function: 
* Why does it return `Result<Option<FileType>>`?
* What is the logical difference between the return values `Ok(None)` and `Err(LoftyError::UnknownFormat)`?
* Why is the return value `Err(LoftyError::UnknownFormat)` from `FileType::from_buffer_inner` transfomed into `Ok(None)`, but internal errors such as `buf_len < 3` use `Err(LoftyError::UnknownFormat)`?
* Would a return type of `Result<FileType>` make more sense?


Serial-ATA:
Thanks for working on this!

The locations of the tests are fine, could you rename the asset to something like `junk.mp3` just to make its purpose a little clearer. Those `a.*` names came from awhile ago, I'll have to update those eventually.

> Why does it return Result<Option<FileType>>?

Since the only error that can occur is `std::io::Error`:

https://github.com/Serial-ATA/lofty-rs/blob/c88cad09b8c0a200eeace10f6e4043fcb2f4c420/src/probe.rs#L138-L142

We just use `Option::or` to set the file type rather than error on no type being found.

> What is the logical difference between the return values Ok(None) and Err(LoftyError::UnknownFormat)?

`Ok(None)` means there were no io errors, but we also couldn't determine the format.

> Why is the return value Err(LoftyError::UnknownFormat) from FileType::from_buffer_inner transfomed into Ok(None), but internal errors such as buf_len < 3 use Err(LoftyError::UnknownFormat)?

That's actually a mistake, the length check shouldn't be there at all. I forgot to remove those when I chose to only return `io::Error`. Good catch. Could you remove the instances of `Err(LoftyError::*)` and just return `Ok(None)` in their place?

src/mp3/header.rs
```diff
@@ -9,6 +9,36 @@ pub(crate) fn verify_frame_sync(frame_sync: [u8; 2]) -> bool {
 	frame_sync[0] == 0xFF && frame_sync[1] >> 5 == 0b111
 }
 
+/// Searches for a frame sync (11 bits with the value 1 like `0b1111_1111_111`) in the input reader.
+/// The search starts at the beginning of the reader and returns the index relative to this beginning.
+/// Only the first match is returned and on no match, [`None`] is returned instead.
+///
+/// Note that the search searches in 8 bit steps, i.e. the first 8 bits need to be byte aligned.
+pub(crate) fn search_for_frame_sync(input: &mut dyn Read) -> Result<Option<u64>> {
```
Serial-ATA, line 17:
```suggestion
pub(crate) fn search_for_frame_sync<R>(input: &mut R) -> Result<Option<u64>>
where
    R: Read,
{
```

src/probe.rs
```diff
@@ -149,41 +149,62 @@ impl<R: Read + Seek> Probe<R> {
...
+					} else if &ident == b"MAC" {
+						Ok(Some(FileType::APE))
+					} else {
+						// potentially some junk bytes are between the ID3 block and the following MP3 block
+						// search for any possible sync bits after the ID3 block
+						self.inner.seek(SeekFrom::Start(position_after_id3_block))?;
+						if let Some(_) = search_for_frame_sync(&mut self.inner)? {
+							Ok(Some(FileType::MP3))
+						} else {
+							Err(LoftyError::UnknownFormat)
```
Serial-ATA, line 199:
```suggestion
							Ok(None)
```

src/probe.rs
```diff
@@ -247,3 +268,35 @@ where
...
+			// start of MP3 frame (not all bytes are shown in this slice)
+			&[
+				0xFF, 0xFB, 0x50, 0xC4, 0x00, 0x03, 0xC0, 0x00, 0x01, 0xA4, 0x00, 0x00, 0x00, 0x20,
+				0x00, 0x00, 0x34, 0x80, 0x00, 0x00, 0x04,
+			],
+		];
+		let data: Vec<u8> = data.into_iter().flatten().cloned().collect();
+		let data = std::io::Cursor::new(&data);
+		let probe = Probe::new(data);
+		let probe = probe.guess_file_type().unwrap();
```
Serial-ATA, line 299:
```suggestion
		let probe = Probe::new(data).guess_file_type().unwrap();
```

This can just be one line :)

src/mp3/read.rs
```diff
@@ -80,17 +80,30 @@ where
...
+				// seek back the length of the temporary header buffer
+				// so that all bytes are included in the search for a frame sync
+				let start_of_search_area =
+					reader.seek(SeekFrom::Current(-1 * header.len() as i64))?;
+				if let Some(first_mp3_frame_start_relative) = search_for_frame_sync(reader)? {
+					let first_mp3_frame_start_absolute =
+						start_of_search_area + first_mp3_frame_start_relative;
+
+					// read the first four bytes of the found frame
+					reader.seek(SeekFrom::Start(first_mp3_frame_start_absolute))?;
```
Serial-ATA, line 94: Aren't we already here?

src/probe.rs
```diff
@@ -149,41 +149,62 @@ impl<R: Read + Seek> Probe<R> {
 
 	#[allow(clippy::shadow_unrelated)]
 	fn guess_inner(&mut self) -> Result<Option<FileType>> {
```
Serial-ATA, line 151:
```suggestion
	fn guess_inner(&mut self) -> std::io::Result<Option<FileType>> {
```

We can just return an `io::Result` now

src/probe.rs
```diff
@@ -149,41 +149,62 @@ impl<R: Read + Seek> Probe<R> {
...
+					// try to guess the file type after the ID3 block by inspecting the first 3 bytes
+					let mut ident = [0; 3];
+					let buf_len = std::io::copy(
+						&mut self.inner.by_ref().take(ident.len() as u64),
+						&mut Cursor::new(&mut ident[..]),
+					)?;
+
+					if buf_len < 3 {
+						Err(LoftyError::UnknownFormat)
+					} else if &ident == b"MAC" {
```
Serial-ATA, line 190:
```suggestion
					if &ident == b"MAC" {
```

We don't need this check anymore

Serial-ATA (changes requested):
Looking good! Just some minor things.

src/mp3/read.rs
localthomas: No, because the call to `search_for_frame_sync(reader)` changes the reader and the first two bytes of the MP3 frame are consumed. The function needs to read the first 11 bits, i.e. the first two bytes, and those would not be included in the follwing `reader.read_u32`.

It could be replaced by `reader.Seek(SeekFrom::Current(-2))`, but I think this would be less readable and robust.

localthomas:
I have updated the pull request according to the code review and also fixed the clippy errors (very good to hve that in CI).

---

Would it be useful to also change the type definition of `Probe::guess_file_type` to `std::io::Result` to make the comment about "All errors that occur within this function are `std::io::Error`." irrelevant?
It would change the public API, so this might not be desirable to do in this pull request.

src/mp3/read.rs
Serial-ATA: Ah, yeah this is definitely the more readable option. Forgot this will leave the reader past the frame sync.

Serial-ATA:
Yeah, I'll save that change for the next major version. This can be included in the next minor release.

Also, could you revert that clippy lint for the negative multiplication? I think that was a lot more readable. You can just throw an `#[allow(clippy::neg_multiply)]` in there.

Serial-ATA (approved):

Serial-ATA:
Thanks!

PR #31, opened 2022-01-23
Added bit depth for all supported formats

sagudev:
Tests are also working, the only thing that is not perfect is WAV as different formats have different ways to store bits per sample.


src/ape/properties.rs
```diff
@@ -147,7 +156,7 @@ where
 	}
 
 	// Unused
```
Serial-ATA, line 158:
```suggestion
```

src/ape/properties.rs
```diff
@@ -203,7 +213,15 @@ where
 	let compression_level = header_first.read_u16::<LittleEndian>()?;
 
 	// Unused
```
Serial-ATA, line 215:
```suggestion
```

src/iff/wav/properties.rs
```diff
@@ -32,6 +32,7 @@ pub struct WavProperties {
 	overall_bitrate: u32,
 	audio_bitrate: u32,
 	sample_rate: u32,
+	bit_depth: u8,
```
Serial-ATA, line 35: This needs a getter

src/ape/properties.rs
```diff
@@ -70,6 +74,11 @@ impl ApeProperties {
 		self.sample_rate
 	}
 
+	/// bits per sample
```
Serial-ATA, line 77:
```suggestion
	/// Bits per sample
```

src/ape/properties.rs
```diff
@@ -203,7 +213,15 @@ where
 	let compression_level = header_first.read_u16::<LittleEndian>()?;
 
 	// Unused
-	let _format_flags = header_first.read_u16::<LittleEndian>()?;
+	let format_flags = header_first.read_u16::<LittleEndian>()?;
+	// https://github.com/fernandotcl/monkeys-audio/blob/5fe956c7e67c13daa80518a4cc7001e9fa185297/src/MACLib/MACLib.h#L74
+	let bit_depth = if (format_flags & 0b1) == 1 {
```
Serial-ATA, line 218:
```suggestion
	let bit_depth = if (format_flags & 0x01) == 1 {
```

src/ape/properties.rs
```diff
@@ -203,7 +213,15 @@ where
...
+		8
+	} else if (format_flags & 0b100) == 4 {
```
Serial-ATA, line 220:
```suggestion
	} else if (format_flags & 0x04) == 4 {
```

Serial-ATA (changes requested):
Thanks! Just a few things.

Also, could you add this to `Mp4Properties`? We are still skipping it:

https://github.com/Serial-ATA/lofty-rs/blob/81bce69d18bdf248d87d0a61a23e5b7f2eb3bdb9/src/mp4/properties.rs#L259

https://github.com/Serial-ATA/lofty-rs/blob/81bce69d18bdf248d87d0a61a23e5b7f2eb3bdb9/src/mp4/properties.rs#L342

src/ape/properties.rs
sagudev: Why? I think it's much more clearer if we keep it in binary format. Especially latter. It highlights the intention.

sagudev:
> Also, could you add this to Mp4Properties?

I just checked and `ffprobe` does not show any bit-depth, which isn't surprising as it is lossy and lossy codecs do not have real bit-depth. All other file formats had bit-depth sown in `ffprobe`.

However, this only applies if m4a is packed with lossy codecs. If it is packed with ALAC, we can get the actual depth. 

For creating testfile I converted `a.flac` to `b.m4a`. But some ALAC data is wrongly calculated (both bitrates), so the test is only checking against bit-depth. But I think fixing those is out of this PR.

The ALAC properties that are returned:
```
Mp4Properties { 
  codec: ALAC, 
  duration: 1.428s, 
  overall_bitrate: 0, 
  audio_bitrate: 1536000, 
  sample_rate: 48000, 
  bit_depth: Some(16), 
  channels: 2 
}
```

src/ogg/flac/properties.rs
```diff
@@ -32,6 +32,7 @@ where
 	let info = stream_info.read_u32::<BigEndian>()?;
 
 	let sample_rate = info >> 12;
+	let bits_per_sample = ((info >> 4) & 0b11111) + 1;
 	let channels = ((info >> 9) & 7) + 1;
```
sagudev, line 36: Here is also one case where binary format is more readable.

src/ape/properties.rs
Serial-ATA: This was just for consistency's sake, since we use hex everywhere else. It's fine though if you think it's clearer.

src/ogg/flac/properties.rs
Serial-ATA: Yeah this one's fine. I was using decimal for the channels anyway.

Serial-ATA (approved):

Serial-ATA:
Thanks!

PR #61, opened 2022-08-03
Fix typo in `InitialKey` key

EmperorJack:
Hi there, I noticed a potential typo in the key used for `InitialKey`. According to the [id3v2.4](https://id3.org/id3v2.4.0-frames) spec, the frame for the initial key should be called `TKEY`, not `TLEY`.

Really enjoying using this crate, thank you!


Serial-ATA:
Oops 😅. Thanks!

PR #64, opened 2022-08-20
ID3v2: populate popularimeter tag from frame value

### LINKED ISSUE #63: ID3v2 popularimeter is not populated
### Reproducer

Rust version: v1.63.0 stable

`cargo.toml`: 
```toml
[package]
name = "lofty-id3-pop"
version = "0.1.0"
edition = "2021"

[dependencies]
lofty = "0.8.0"
```

`src/main.rs`:
```rust
fn main() {
    // Contents of the file
    let mut example_bytes = vec![
        // ID3 header - https://id3.org/id3v2.4.0-structure section 3
        b'I', b'D', b'3', 0x04, 0x00, 0b00000000, 0, 0, 0, 118,
        // ID3 POPM frame - https://en.wikipedia.org/wiki/ID3#ID3v2_star_rating_tag_issue
        b'P', b'O', b'P', b'M', 0, 0, 0, 35, 0x00, 0x00, b'W', b'i', b'n', b'd', b'o', b'w', b's',
        b' ', b'M', b'e', b'd', b'i', b'a', b' ', b'P', b'l', b'a', b'y', b'e', b'r', b' ', b'9',
        b' ', b'S', b'e', b'r', b'i', b'e', b's', 0x00, 196, 0x00, 0x00, 0x00, 0x00,
    ];
    // Pad to at least 128 bytes to support the negative seek during ID3v1 header parsing
    // - https://github.com/Serial-ATA/lofty-rs/tree/1b546700d552caeb970a1751862b8b5a2585fe9c/src/mpeg/read.rs#L100
    // - https://github.com/Serial-ATA/lofty-rs/tree/1b546700d552caeb970a1751862b8b5a2585fe9c/src/id3/mod.rs#L54
    example_bytes.append(&mut vec![0u8; 128 - 55]);

    // Write the file to disk
    let path = std::env::temp_dir().join("minimal_example.mp3");
    std::fs::write(&path, example_bytes).unwrap();

    // Parse the file for ID3 tags
    let info = lofty::read_from_path(&path, false).unwrap();
    for tag in info.tags() {
        if let Some(pop) = tag.get_string(&lofty::ItemKey::Popularimeter) {
            println!("Found popularimeter: {pop}");
            return;
        }
    }
    panic!("No popularimeter tag found");
}
```

### Summary

When parsing files with ID3v2 tags, such as MP3s, the Popularimeter tag is never populated. 

### Expected behavior

Using the supplied example:

Expected:
```sh
> cargo run
Found popularimeter: 196
```

Actual:
```sh
> cargo run
thread 'main' panicked at 'No popularimeter tag found', src/main.rs:28:5
```

### Assets

_No response_

maneac:
Use the `rating` field as the value of the Popularimeter tags when handling ID3v2 tags.

closes #63

### DIFF
--- a/src/id3/v2/items/popularimeter.rs
+++ b/src/id3/v2/items/popularimeter.rs
@@ -61,7 +61,7 @@ impl Popularimeter {
 				1 => pop.rating = *chunk.first().unwrap_or(&0),
 				2 => chunk.iter().for_each(|&b| {
 					pop.counter <<= 8;
-					pop.counter |= u64::from(b);
+					pop.counter += u64::from(b);
 				}),
 				_ => unreachable!(),
 			}
--- a/src/id3/v2/items/popularimeter.rs
+++ b/src/id3/v2/items/popularimeter.rs
@@ -1,3 +1,4 @@
+use crate::error::LoftyError;
 use crate::util::text::{encode_text, TextEncoding};
 
 use std::hash::{Hash, Hasher};
@@ -44,6 +45,29 @@ impl Popularimeter {
 		}
 
 		content
+	}
+
+	/// Convert ID3v2 POPM frame bytes into a [`Popularimeter`].
+	pub fn from_bytes(bytes: &[u8]) -> Result<Self, LoftyError> {
+		let mut pop = Self {
+			email: String::new(),
+			rating: 0,
+			counter: 0,
+		};
+
+		for (idx, chunk) in bytes.splitn(3, |b| b.eq(&0)).enumerate() {
+			match idx {
+				0 => pop.email = String::from_utf8(chunk.to_vec())?,
+				1 => pop.rating = *chunk.first().unwrap_or(&0),
+				2 => chunk.iter().for_each(|&b| {
+					pop.counter <<= 8;
+					pop.counter |= u64::from(b);
+				}),
+				_ => unreachable!(),
+			}
+		}
+
+		Ok(pop)
 	}
 }
 
--- a/src/id3/v2/items/popularimeter.rs
+++ b/src/id3/v2/items/popularimeter.rs
@@ -1,6 +1,7 @@
-use crate::error::LoftyError;
-use crate::util::text::{encode_text, TextEncoding};
+use crate::error::Result;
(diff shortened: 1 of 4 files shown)


Serial-ATA:
Thanks! It might be better to use `Popularimeter::as_bytes` and store it as `ItemValue::Binary` to preserve the other fields, just in case they're written.

Could you also document the behavior [here](https://github.com/Serial-ATA/lofty-rs/blob/cffc8182e13e908cc707c2d5bd9246e12caec67a/src/id3/v2/tag.rs#L87)? 🙂

maneac:
Hey, I've updated the PR to now use the binary encoding to preserve all fields of the tag.
This highlighted further issues when converting between the formats, which should now also be resolved.

Documentation has been updated as suggested. I believe the existing documentation elsewhere already covers the expected behaviour, but I'm happy to update further if needed :)

src/id3/v2/items/popularimeter.rs
```diff
@@ -46,6 +47,29 @@ impl Popularimeter {
 
 		content
 	}
+
+	/// Convert ID3v2 POPM frame bytes into a [`Popularimeter`].
+	pub fn from_bytes(bytes: &[u8]) -> Result<Self, LoftyError> {
```
Serial-ATA, line 52: You can just copy the logic from here (without wrapping it in `FrameValue`):
https://github.com/Serial-ATA/lofty-rs/blob/f25ed9b6c7c920263c57ab327d023e517ae0e16a/src/id3/v2/frame/content.rs#L165-L186

And then change this:
https://github.com/Serial-ATA/lofty-rs/blob/f25ed9b6c7c920263c57ab327d023e517ae0e16a/src/id3/v2/frame/content.rs#L37

to:
```rust
"POPM" => Some(FrameValue::Popularimeter(Popularimeter::from_bytes(content)?)),
```

Be sure to update your branch, I just edited `parse_popularimeter` :slightly_smiling_face:.

Serial-ATA:
Hey @maneac, I just went ahead and made the changes to make sure this gets in `0.9.0`. Thanks for your work on this! :)

PR #65, opened 2022-08-31
Add example of reading album artist to tag reader

cmdcolin:
Just thought this can help clarify how to read other types of tags in the tag reader example :)

Let me know if this helps

### DIFF
--- a/examples/tag_reader.rs
+++ b/examples/tag_reader.rs
@@ -1,3 +1,4 @@
+use lofty::ItemKey::AlbumArtist;
 use lofty::{Accessor, AudioFile, Probe};
 use std::path::Path;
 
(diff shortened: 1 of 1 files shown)


examples/tag_reader.rs
```diff
@@ -28,6 +29,12 @@ fn main() {
 	println!("Album: {}", tag.album().unwrap_or("None"));
 	println!("Genre: {}", tag.genre().unwrap_or("None"));
 
+	// import keys from https://docs.rs/lofty/latest/lofty/enum.ItemKey.html
+	println!(
+		"Album Artist: {}",
+		tag.get_string(&AlbumArtist).unwrap_or("None")
```
Serial-ATA, line 35: Thanks! Could use you the full path here instead of importing the variant, just to be extra clear? 🙂

Serial-ATA:
Thanks! This is nice to have. :slightly_smiling_face:

PR #98, opened 2023-01-05
mp4: Add new item key AppleXID

uklotzde:
This atom is only contained in files bought from the iTunes store, but very common for MP4 files and it contains valuable information.

The format and contents are defined in the development guide: https://www.apple.com/nz/itunes/lp-and-extras/docs/Development_Guide.pdf


uklotzde:
I have simplified the naming.

PR #102, opened 2023-01-06
Reduce memory allocations for id3v2::FrameID

uklotzde:
The next step towards more efficient memory management. Most of the predefined string constants have a `'static` lifetime.

### DIFF
--- a/src/id3/v2/frame/id.rs
+++ b/src/id3/v2/frame/id.rs
@@ -1,33 +1,35 @@
+use std::borrow::Cow;
+
 use crate::error::{ID3v2Error, ID3v2ErrorKind, LoftyError, Result};
 use crate::tag::item::ItemKey;
 use crate::tag::TagType;
 
 /// An `ID3v2` frame ID
 #[derive(PartialEq, Clone, Debug, Eq, Hash)]
-pub enum FrameID {
+pub enum FrameID<'a> {
 	/// A valid `ID3v2.3/4` frame
-	Valid(String),
+	Valid(Cow<'a, str>),
 	/// When an `ID3v2.2` key couldn't be upgraded
 	///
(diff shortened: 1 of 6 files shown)


src/id3/v2/frame/id.rs
```diff
@@ -1,45 +1,47 @@
...
+impl<'a> FrameID<'a> {
 	/// Attempts to create a `FrameID` from an ID string
 	///
 	/// # Errors
 	///
 	/// * `id` contains invalid characters (must be 'A'..='Z' and '0'..='9')
 	/// * `id` is an invalid length (must be 3 or 4)
-	pub fn new(id: &str) -> Result<Self> {
-		Self::verify_id(id)?;
+	pub fn new(id: Cow<'a, str>) -> Result<Self> {
```
Serial-ATA, line 27:
```suggestion
	pub fn new<I>(id: I) -> Result<Self>
	where
		I: Into<Cow<'a, str>>,
	{
```

I think this should be generic, I'd rather callers not have to deal with `Cow`s outside of return types.

src/id3/v2/frame/mod.rs
```diff
@@ -73,16 +73,16 @@ impl Frame {
 	///
 	/// * `id` is less than 3 or greater than 4 bytes
 	/// * `id` contains non-ascii characters
-	pub fn new(id: &str, value: FrameValue, flags: FrameFlags) -> Result<Self> {
+	pub fn new(id: Cow<'a, str>, value: FrameValue, flags: FrameFlags) -> Result<Self> {
```
Serial-ATA, line 76: Same as above

Serial-ATA (changes requested):

src/id3/v2/frame/id.rs
uklotzde: Agreed. I found a way to avoid code bloat by monomorphization.

PR #112, opened 2023-01-09
id3v2: Use "XXX" if the actual language is unknown

uklotzde:
See my comment directly in the code.


Serial-ATA:
Thanks! Good to know there's a placeholder value for this.

PR #115, opened 2023-01-10
mp4: Implement IntoIterator for Ilst

uklotzde:
This is needed for merging tags.

### DIFF
--- a/src/mp4/ilst/mod.rs
+++ b/src/mp4/ilst/mod.rs
@@ -83,11 +83,6 @@ pub struct Ilst {
 }
 
 impl Ilst {
-	/// Returns all of the tag's atoms
-	pub fn atoms(&self) -> &[Atom<'static>] {
-		&self.atoms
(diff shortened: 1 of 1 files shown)


src/mp4/ilst/mod.rs
```diff
@@ -84,6 +84,11 @@ pub struct Ilst {
...
 impl Ilst {
 	/// Returns all of the tag's atoms
+	///
+	/// # Deprecated
+	///
+	/// Use one of the [`IntoIterator`] implementations for accessing all atoms.
+	#[deprecated]
 	pub fn atoms(&self) -> &[Atom<'static>] {
 		&self.atoms
 	}
```
Serial-ATA, line 94:
```suggestion
```
No need to mark this as deprecated, it can just be removed.

uklotzde: Amended and force pushed.

PR #116, opened 2023-01-10
Use `#[must_use] pub const fn new(...)` whenever possible

uklotzde:


Serial-ATA:
Thanks, been meaning to do this.

PR #119, opened 2023-01-11
Fix base64 deprecation warnings

uklotzde:


Serial-ATA:
Thanks, just blindly merged #111.

PR #121, opened 2023-01-11
id3v2: Distinguish COMM frames by content descriptor

uklotzde:
Plain comments are supposed to have an empty description aka *content descriptor*. Only those must appear as `ItemKey::Comment`, otherwise this would cause confusion and inconsistencies during the read/write round trip.

### TODO
- [x] How to filter the COMM frames such that only those with an empty content descriptor appear under `ItemKey::Comment`?
- [x] ~~Test that unsupported frames with a non-empty content descriptor are preserved during a read/write round trip.~~ Out of scope, nothing to do here. Conversions from `ID3v2Tag` to `Tag` are expected to be lossy and applications are responsible for preserving unmapped MP3 metadata frames. All operations directly on `ID3v2Tag` are unaffected by the changes.

### DIFF
--- a/src/id3/v2/frame/mod.rs
+++ b/src/id3/v2/frame/mod.rs
@@ -20,12 +20,24 @@ use crate::id3::v2::items::popularimeter::Popularimeter;
 use std::convert::{TryFrom, TryInto};
 use std::hash::{Hash, Hasher};
 
-// <https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-structure.html>
-//
-// > The three byte language field, present in several frames, is used to describe
-// > the language of the frame’s content, according to ISO-639-2 [ISO-639-2].
-// > The language should be represented in lower case. If the language is not known
-// > the string “XXX” should be used.
+/// Empty content descriptor in text frame
+///
+/// Unspecific [`LanguageFrame`]s and [`EncodedTextFrame`] frames
+/// are supposed to have an empty content descriptor. Only those
+/// are currently supported as [`TagItem`]s to avoid ambiguities
+/// and to prevent inconsistencies when writing them.
+pub(super) const fn empty_content_descriptor() -> String {
+	String::new()
+}
+
+/// Unknown language-aware text frame
+///
+/// <https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-structure.html>
+///
+/// > The three byte language field, present in several frames, is used to describe
+/// > the language of the frame’s content, according to ISO-639-2 [ISO-639-2].
+/// > The language should be represented in lower case. If the language is not known
+/// > the string “XXX” should be used.
 pub(super) const UNKNOWN_LANGUAGE: [u8; 3] = *b"XXX";
 
 // TODO: Messy module, rough conversions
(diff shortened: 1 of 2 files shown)


src/id3/v2/tag.rs
```diff
@@ -266,16 +271,11 @@ impl ID3v2Tag {
 		})
 	}
 
-	/// Returns all `COMM` frames
+	/// Returns all `COMM` frames with an empty content descriptor
```
Serial-ATA, line 274: Why should this be restricted to empty descriptors only? An ID3v2 tag shouldn't have any comments with duplicate descriptors according to spec anyway. It's definitely possible to create them and they will be written, but I don't think that should be encouraged.

uklotzde: Those custom comments with non-empty content descriptors cause lots of issues. Think about all the possible round trips and you will notice how they cause inconsistencies and unexpected behavior.

If you are not able to distinguish them the restrict access to those with an empty content descriptor. Experience from Mixxx development and from seeing lots of different files in the wild.

TagLib: <https://github.com/taglib/taglib/blob/a31356e330674640a07bef7d71d08242cae8e9bf/taglib/mpeg/id3v2/id3v2tag.cpp#L175>
Mixxx: <https://github.com/mixxxdj/mixxx/blob/cce89d09f3a546f54af2c2d473a848f9cf6eee26/src/track/taglib/trackmetadata_id3v2.cpp#L697>

...and probably many others also.

uklotzde: TagLib picks the first non-empty content descriptor as a fallback, which is bad because it can cause unexpected behavior. After writing the comment the original frame with the non-empty content descriptor remains in the file, probably even as the first one. Depending on whether other applications are handling the distinction by content descriptor correctly or not users will complain about comments not being written... Have fun.

Being strict and consistent is the only chance to avoid chaos.

Serial-ATA: Thanks, I see how that could be an issue.

Serial-ATA:
For the TODO, the `ID3v2Tag` -> `Tag` conversion could easily be changed to only accept empty descriptors here:
https://github.com/Serial-ATA/lofty-rs/blob/90a7683322b83d5d81e2a5223d2b06c4bd55c919/src/id3/v2/tag.rs#L579

The real issue is how should comments *with* a descriptor be stored? Currently, the descriptor just gets discarded, which definitely isn't ideal. Might have to take a similar route to the popularimeter, adding a new `ItemKey` and converting the `LanguageFrame` to binary.

src/id3/v2/frame/mod.rs
```diff
@@ -20,12 +20,24 @@ use crate::id3::v2::items::popularimeter::Popularimeter;
...
-// > the string “XXX” should be used.
+/// Empty content descriptor in text frame
+///
+/// Unspecific [`LanguageFrame`]s and [`EncodedTextFrame`] frames
+/// are supposed to have an empty content descriptor. Only those
+/// are currently supported as [`TagItem`]s to avoid ambiguities
+/// and to prevent inconsistencies when writing them.
+pub(super) const fn empty_content_descriptor() -> String {
+	String::new()
+}
```
Serial-ATA, line 31:
```suggestion
pub(super) const EMPTY_CONTENT_DESCRIPTOR: String = String::new();
```
Does this have to be a const fn?

uklotzde: Yes, because a constant would require explicit `.clone()` operations in the code.

Serial-ATA: You're able to swap it out with a constant in this case without clones.

uklotzde: Your totally right. My assumptions were wrong.

PR #132, opened 2023-01-18
Split and rejoin tags for read/modify/write round trips

uklotzde:
Adds a new trait `SplitAndRejoinTag` with methods that are needed for safely and efficiently implementing read/modify/write round trips when synchronizing metadata:

- `fn split_tag(&mut self) -> Tag;`
- `fn rejoin_tag(&mut self);`

The central function from my code base that accomplishes the task demonstrates how they are supposed to be used:

```rust
fn split_export_rejoin_track_to_tag<T>(
    tag_repr: &mut T,
    config: &ExportTrackConfig,
    track: &mut Track,
    edit_embedded_artwork_image: Option<EditEmbeddedArtworkImage>,
) where
    T: SplitAndRejoinTag,
{
    // Split the generic tag contents from the underlying representation.
    // The remainder will remain untouched while modifying the generic tag.
    let mut tag = tag_repr.split_tag();
    // Export the metadata of the given track into the generic tag, i.e.
    // add, modify, or delete items and pictures accordingly.
    export_track_to_tag(&mut tag, config, track, edit_embedded_artwork_image);
    // Rejoin the generic tag contents back into the remainder of the underlying
    // representation.
    tag_repr.rejoin_tag(tag);
    // Depending on `T` some post-processing might be required in the outer context
    // to update contents in `tag_repr` that are not (yet) supported by the generic
    // `lofty::Tag` representation.
}
```

~~Ideally the task would be implemented using a type-state-pattern that consumes its arguments to prevent mutating the remainder between split and rejoin. I am considering this alternative, stay tuned...~~ Didn't turn out as helpful. I left a comment in the docs in case anyone has the same idea.

### DIFF
--- a/src/mp4/ilst/atom.rs
+++ b/src/mp4/ilst/atom.rs
@@ -12,10 +12,10 @@ pub(super) enum AtomDataStorage {
 }
 
 impl AtomDataStorage {
-	pub(super) fn take_first(self) -> AtomData {
+	pub(super) fn first_mut(&mut self) -> &mut AtomData {
 		match self {
 			AtomDataStorage::Single(val) => val,
-			AtomDataStorage::Multiple(mut data) => data.swap_remove(0),
+			AtomDataStorage::Multiple(data) => data.first_mut().expect("not empty"),
 		}
 	}
 }
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -233,6 +233,30 @@ pub trait TagExt: Accessor + Into<Tag> + Sized {
 	fn clear(&mut self);
 }
 
+/// Split and rejoin tags.
+///
+/// Useful and required for implementing read/modify/write round trips.
+pub trait SplitAndRejoinTag {
+	/// Extract and split generic contents into a [`Tag`].
+	///
+	/// Leaves the remainder that cannot be represented in the
+	/// resulting tag in `self`. This is useful if the modified [`Tag`]
+	/// is rejoined later using [`SplitAndRejoinTag::rejoin_tag`].
+	// NOTE: Using the "typestate pattern" (http://cliffle.com/blog/rust-typestate/)
+	// to represent the intermediate state turned out as less flexible
+	// and useful than expected.
+	fn split_tag(&mut self) -> Tag;
+
+	/// Rejoin a [`Tag`].
+	///
+	/// Rejoin a tag that has previously been obtained with
+	/// [`SplitAndRejoinTag::split_tag`].
+	///
+	/// Restores the original representation merged with the contents of
+	/// `tag` for further processing, e.g. writing back into a file.
+	fn rejoin_tag(&mut self, tag: Tag);
+}
+
 // TODO: https://github.com/rust-lang/rust/issues/59359
 pub(crate) trait SeekStreamLen: std::io::Seek {
 	fn stream_len(&mut self) -> crate::error::Result<u64> {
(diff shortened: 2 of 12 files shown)


src/mp4/ilst/atom.rs
```diff
@@ -12,10 +12,10 @@ pub(super) enum AtomDataStorage {
 }
 
 impl AtomDataStorage {
-	pub(super) fn take_first(self) -> AtomData {
+	pub(super) fn first_mut(&mut self) -> &mut AtomData {
 		match self {
 			AtomDataStorage::Single(val) => val,
-			AtomDataStorage::Multiple(mut data) => data.swap_remove(0),
+			AtomDataStorage::Multiple(data) => data.first_mut().expect("not emtpy"),
```
Serial-ATA, line 18:
```suggestion
			AtomDataStorage::Multiple(data) => data.first_mut().expect("not empty"),
```

src/traits.rs
```diff
@@ -233,6 +233,33 @@ pub trait TagExt: Accessor + Into<Tag> + Sized {
...
+	///
+	/// Leaves the remainder that cannot be represented in the
+	/// resulting tag in `self`. This is useful if the modified [`Tag`]
+	/// is rejoined later using [`SplitAndRejoinTag::rejoin_tag`].
+	///
+	/// This operation leaves a partial tag behind that should only be
+	/// used again after rejoining a tag. Using the
+	/// [typestate pattern](http://cliffle.com/blog/rust-typestate/)
+	/// to represent the intermediate state turned out as less flexible
+	/// and useful than expected.
```
Serial-ATA, line 250:
```suggestion
	// NOTE: Using the "typestate pattern" (http://cliffle.com/blog/rust-typestate/)
	// to represent the intermediate state turned out as less flexible
	// and useful than expected.
```
This is only really useful to anyone trying to implement this in the future, it should just be a normal comment.

src/traits.rs
```diff
@@ -233,6 +233,33 @@ pub trait TagExt: Accessor + Into<Tag> + Sized {
...
+	fn split_tag(&mut self) -> Tag;
+
+	/// Rejoin a [`Tag`].
+	///
+	/// Rejoin a a tag that has previously been obtained with
```
Serial-ATA, line 255:
```suggestion
	/// Rejoin a tag that has previously been obtained with
```

src/traits.rs
```diff
@@ -233,6 +233,33 @@ pub trait TagExt: Accessor + Into<Tag> + Sized {
 	fn clear(&mut self);
 }
 
+/// Split and rejoin tags.
+///
+/// Useful and required for implementing read/modify/write round trips.
+pub trait SplitAndRejoinTag {
```
Serial-ATA, line 239: Can you add an `# Examples` section to the methods? I try to do that with all new public methods.

Serial-ATA (changes requested):
This is a good idea! Thanks for implementing it. I just have some doc changes.

src/traits.rs
uklotzde: I already thouught about that when adding my code example here. Was just lazy ;)

uklotzde: Since you always apply these operations in pairs the example should be added to the trait.

Serial-ATA:
Having thought about it, I think `rejoin_tag` should actually be `merge_tag`. Since you could really use it to add any tag into another.

PR #137, opened 2023-01-30
id3v2/mp4/ogg: Add tests for multi-valued item conversion

uklotzde:
The support for multi-valued items in ID3v2 tags was very limited as I noticed while fixing an issue in the context of #103:

- Extend support for multi-valued field in ID3v2 tags, i.e. not only `TrackArtist`.
- Warn when replacing and dropping multi-valued items in ID3v2 tags unintentionally.
- Add tests for conversion round trips with multi-valued items.

This is probably not the most elegant solution, but at least a start accompanied by tests.

### DIFF
--- a/src/id3/v2/tag.rs
+++ b/src/id3/v2/tag.rs
@@ -684,7 +684,7 @@ impl SplitAndMergeTag for ID3v2Tag {
 	}
 
 	fn merge_tag(&mut self, mut tag: Tag) {
-		fn join_items(tag: &mut Tag, key: &ItemKey) -> Option<String> {
+		fn join_text_items(tag: &mut Tag, key: &ItemKey) -> Option<String> {
 			let mut iter = tag.take_strings(key);
 			iter.next().map(|first| {
 				// Use the length of the first string for estimating the capacity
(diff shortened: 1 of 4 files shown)


src/id3/v2/tag.rs
```diff
@@ -708,19 +708,65 @@ impl SplitAndMergeTag for ID3v2Tag {
...
+				self.remove(frame_id);
+			}
 		}
 
+		let comment_frame_id = ItemKey::Comment
+			.map_key(TagType::ID3v2, false)
+			.expect("valid frame id");
+		if let Some(text) = join_text_items(&mut tag, &ItemKey::Comment) {
+			let frame = Frame {
+				id: FrameID::Valid(Cow::Borrowed(comment_frame_id)),
```
Serial-ATA, line 747:
```suggestion
		if let Some(text) = join_text_items(&mut tag, &ItemKey::Comment) {
			let frame = Frame {
				id: FrameID::Valid(Cow::Borrowed(COMMENT_FRAME_ID)),
```

src/id3/v2/tag.rs
```diff
@@ -708,19 +708,53 @@ impl SplitAndMergeTag for ID3v2Tag {
...
+				};
+				self.insert(frame);
+			} else {
+				self.remove(frame_id);
+			}
 		}
 
+		if let Some(text) = join_text_items(&mut tag, &ItemKey::Comment) {
+			// The first comment frame is either replaced or added.
+			debug_assert!(self.comments().count() <= 1);
```
Serial-ATA, line 744: This assertion isn't always valid, since you can merge a Tag with an ID3v2Tag that already has comments inside of it.

Also, you should just take the first comment instead of concatenating all of them into one. I feel like the former is less surprising.

uklotzde: The merge operation should only be invoked on the untouched, remaining part after splitting a tag. The remaining part does not contain any of those item keys, otherwise the results are undefined (as documented).

This is not supposed to be a universal merge operation. Otherwise the could would need to be less efficient and more complex.

Serial-ATA: It wouldn't be UB in this case, as `ID3v2Tag::insert` and `ID3v2Tag::set_comment` are being used. Those both will remove any collisions prior to insertion, so it would be safe to merge even if the tag has been modified.

uklotzde: I have revisited the ID3v2 implementation and eliminated some inconsistencies. Debug assertions indicate situations were undefined behavior occurs when the preconditions of this operation are violated.

There was also a bug: Simply appending the picture frames was wrong, because it could violate the ID3v2 restrictions. Deduplication during the key-to-frame mapping by using `insert()` is required here and not covered by the preconditions.

uklotzde:
I have rebased the branch.

Please see my comments and the debug assertions that document the design decisions. I have also checked that the implementation is consistent with the other formats as much as possible. All implementations make assumptions about the context and its preconditions, i.e. undefined behavior if the merge target contains information that is supposed to be absent after the `split_tag()` operation.

uklotzde:
I am planning to prevent the UB at runtime by applying the typestate pattern. This will be resolved by a follow-up PR.

Serial-ATA:
Sorry for the delay on this, I'll take a look in the next 2 days.

PR #142, opened 2023-02-11
ID3v2: Fix conversion of user defined frames when using `Tag` writing interface

### LINKED ISSUE #140: [Bug]: Writing `ID3v2` tag to mp3 file containing replay gain frames generated by `r128gain` tool fails
### Reproducer

Using lofty `0.11` I tried this code:

```rust
fn tag_test(path: impl AsRef<Path>) -> Result<(), LoftyError> {
    let mut tagged_file = read_from_path(&path)?;
    if let Some(tag) = tagged_file.tag_mut(TagType::ID3v2) {
        tag.set_title("foo".into());
        tag.save_to_path(&path)?;
    }
    Ok(())
}
```

Which doesn't work for a file with the following `TXXX` frames (as reported by the `mid3v2` command line tool, I don't think the specific values matter too much as I could reproduce this on multiple files with differing replay gains):
```
TXXX=REPLAYGAIN_TRACK_GAIN=0.50 dB
TXXX=REPLAYGAIN_TRACK_PEAK=0.479950
```

It fails with the error:
```
ID3v2(ID3v2: BadFrame("REPLAYGAIN_TRACK_GAIN", "Text"))
```

After removing these frames lofty is able to write the tag to the files again. Letting the `r128gain` tool write out the replay gain values again makes lofty fail to write the tag again. I haven't tried with another replay gain analysis tool, so it could be the case that the tags written by `r128gain` are wrong but I think this is unlikely as I've not had any other issues with it nor can I find issues about it writing wrong tags.

### Summary

Lofty fails to write back `ID3v2` tag if it contains replay gain frames written by `r128gain`

### Expected behavior

For lofty to successfully write tags containing replay gain tags.

### Assets

I've analyzed this file: [full_test.mp3](https://github.com/Serial-ATA/lofty-rs/blob/main/tests/files/assets/minimal/full_test.mp3)
with the `r128gain` tool which successfully reproduces the bug. Here's the file, uploaded as a `jpg` since github doesn't accept `mp3` files, so just change the extension to `mp3` again: full_test_rg128gain

Serial-ATA:
fixes #140

### DIFF
--- a/src/id3/v2/frame/mod.rs
+++ b/src/id3/v2/frame/mod.rs
@@ -9,7 +9,7 @@ use crate::id3::v2::items::language_frame::LanguageFrame;
 use crate::id3::v2::util::upgrade::{upgrade_v2, upgrade_v3};
 use crate::id3::v2::ID3v2Version;
 use crate::picture::Picture;
-use crate::tag::item::{ItemKey, ItemValue, TagItem};
+use crate::tag::item::{ItemValue, TagItem};
 use crate::tag::TagType;
 use crate::util::text::{encode_text, TextEncoding};
 use id::FrameID;
@@ -362,7 +362,7 @@ impl From<TagItem> for Option<Frame<'static>> {
 }
 
 pub(crate) struct FrameRef<'a> {
-	pub id: &'a str,
+	pub id: FrameID<'a>,
 	pub value: Cow<'a, FrameValue>,
 	pub flags: FrameFlags,
 }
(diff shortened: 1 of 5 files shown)


src/id3/v2/frame/mod.rs
```diff
@@ -371,7 +371,7 @@ impl<'a> Frame<'a> {
 	pub(crate) fn as_opt_ref(&'a self) -> Option<FrameRef<'a>> {
 		if let FrameID::Valid(id) = &self.id {
 			Some(FrameRef {
-				id,
+				id: FrameID::Valid(id.clone()),
```
uklotzde, line 374: `Cow::Borrowed(&id)` avoids cloning

sandersantema:
Works perfectly using the code and test file I provided in #140 :partying_face: thanks for fixing!

Serial-ATA:
Awesome, thanks for confirming! :slightly_smiling_face:

PR #143, opened 2023-02-12
v0.12.0: Separate SplitAndMergeTag into SplitTag/MergeTag for compile-time safety

uklotzde:
This is the follow-up for #137 to avoid unexpected/undefined behavior at runtime.

Finally, the typestate pattern is in place. Separation into 2 traits was the missing link.


uklotzde:
Finished. Only marked as _Draft_ because it is based on #137 which has not been merged yet and might still change.

Merging this PR instead of #137 would also be an option.

Serial-ATA:
This looks good after a rebase. Thanks!

PR #149, opened 2023-02-20
ID3v2: Fix about number pair of track and disk

### LINKED ISSUE #145: ID3v2 track & track_total frames are written incorrectly (disk and disk_total probably too)
### Reproducer

I tried this code:

```rust
	let mut tagged_file = Probe::open(&opt.path)
		.expect("ERROR: Bad path provided!")
		.read()
		.expect("ERROR: Failed to read file!");

	let tag = tagged_file.primary_tag_mut().unwrap();
        tag.set_track(9);
        tag.set_track_total(13);
        tag.save_to_path(&opt.path).unwrap()
```

### Summary

Track numbers in id3v2 are stored in the TRCK frame, encoded like this `{track_number}/{track_total}`. Lofty-rs tag currently writes track and track_total as two separate TRCK frames. It looks like this in a hex editor: `TRCK.......9TRCK.......13`

### Suggested fix:
in `src/id3/v2/tag.rs` on line 763 in tag_frames add special handling for these frames.

### Expected behavior

Lofty writing the TRCK frame correctly. Like this: `TRCK.......9/13`

### Assets

ID3v2 Spec (ID3 website is currently broken):
https://web.archive.org/web/20191202011405/http://id3.org/id3v2.4.0-frames

mikanbako:
This PR fixes #145.

TODOs

- [x] Rebase to the main branch for conflict.
- [x] Fix id3::v2::tag::tag_frames() that called from tag::Tag::save_to().
- [x] Add tests about frame values.
- [x] Improve parsing number pair.

### DIFF
--- a/src/id3/v2/tag.rs
+++ b/src/id3/v2/tag.rs
@@ -629,32 +629,26 @@ impl SplitAndMergeTag for ID3v2Tag {
 			let mut split =
 				content.splitn(2, &[V4_MULTI_VALUE_SEPARATOR, NUMBER_PAIR_SEPARATOR][..]);
 
-			let number_source = split.next()?;
-			let number = parse_number(number_source)?;
-			let total_source = split.next();
-			let total = total_source.and_then(parse_number);
-
-			let is_valid = if let Some(total_source) = total_source {
-				total.is_some() && content.len() == number_source.len() + 1 + total_source.len()
-			} else {
-				content.len() == number_source.len()
-			};
-
-			if is_valid {
-				tag.items.push(TagItem::new(
-					number_key,
-					ItemValue::Text(number.to_string()),
-				));
-
-				if let Some(total) = total {
-					tag.items
-						.push(TagItem::new(total_key, ItemValue::Text(total.to_string())))
-				}
-
-				Some(())
+			let number = parse_number(split.next()?)?;
+			let total = if let Some(total_source) = split.next() {
+				Some(parse_number(total_source)?)
 			} else {
 				None
+			};
+			debug_assert!(split.next().is_none());
+
+			debug_assert!(!number.is_empty());
+			tag.items.push(TagItem::new(
+				number_key,
+				ItemValue::Text(number.to_string()),
+			));
+			if let Some(total) = total {
+				debug_assert!(!total.is_empty());
+				tag.items
+					.push(TagItem::new(total_key, ItemValue::Text(total.to_string())))
 			}
+
+			Some(())
 		}
 
 		let mut tag = Tag::new(TagType::ID3v2);
--- a/src/id3/v2/tag.rs
+++ b/src/id3/v2/tag.rs
@@ -362,7 +362,14 @@ impl Accessor for ID3v2Tag {
 	}
 
 	fn set_track(&mut self, value: u32) {
-		self.insert(Frame::text(Cow::Borrowed("TRCK"), value.to_string()));
+		let existing_track_total = self.track_total();
+
+		let trck_content = match existing_track_total {
+			Some(track_total) => format!("{value}/{track_total}"),
+			None => value.to_string(),
+		};
+
+		self.insert(Frame::text(Cow::Borrowed("TRCK"), trck_content));
 	}
 
 	fn remove_track(&mut self) {
@@ -374,11 +381,11 @@ impl Accessor for ID3v2Tag {
 	}
 
 	fn set_track_total(&mut self, value: u32) {
-		let current_track = self.split_num_pair("TRCK").0.unwrap_or(1);
+		let track_number = self.track().unwrap_or(1);
 
 		self.insert(Frame::text(
 			Cow::Borrowed("TRCK"),
-			format!("{current_track}/{value}"),
+			format!("{track_number}/{value}"),
 		));
 	}
 
@@ -1633,5 +1640,53 @@ mod tests {
 		let reparsed = crate::id3::v2::read::parse_id3v2(&mut reader, header).unwrap();
 
 		assert_eq!(id3v2, reparsed);
+	}
+
+	#[test]
+	fn set_track() {
+		let mut id3v2 = ID3v2Tag::default();
+		let track = 1;
+
+		id3v2.set_track(track);
+
+		assert_eq!(id3v2.track().unwrap(), track);
+		assert!(id3v2.track_total().is_none());
+	}
+
+	#[test]
+	fn set_track_total() {
+		let mut id3v2 = ID3v2Tag::default();
+		let track_total = 2;
+
+		id3v2.set_track_total(track_total);
+
+		assert_eq!(id3v2.track().unwrap(), 1);
+		assert_eq!(id3v2.track_total().unwrap(), track_total);
+	}
+
+	#[test]
+	fn set_track_and_track_total() {
+		let mut id3v2 = ID3v2Tag::default();
+		let track = 1;
+		let track_total = 2;
+
+		id3v2.set_track(track);
+		id3v2.set_track_total(track_total);
+
+		assert_eq!(id3v2.track().unwrap(), track);
+		assert_eq!(id3v2.track_total().unwrap(), track_total);
+	}
+
+	#[test]
+	fn set_track_total_and_track() {
+		let mut id3v2 = ID3v2Tag::default();
+		let track_total = 2;
+		let track = 1;
+
+		id3v2.set_track_total(track_total);
+		id3v2.set_track(track);
+
+		assert_eq!(id3v2.track_total().unwrap(), track_total);
+		assert_eq!(id3v2.track().unwrap(), track);
 	}
 }
--- a/src/id3/v2/tag.rs
+++ b/src/id3/v2/tag.rs
@@ -14,6 +14,7 @@ use crate::util::text::TextEncoding;
 
 use std::borrow::Cow;
 use std::convert::TryInto;
+use std::fmt::Display;
 use std::fs::{File, OpenOptions};
 use std::io::Write;
 use std::path::Path;
@@ -24,6 +25,17 @@ const COMMENT_FRAME_ID: &str = "COMM";
 
 const V4_MULTI_VALUE_SEPARATOR: char = '\0';
 const NUMBER_PAIR_SEPARATOR: char = '/';
+
+// This is used as the default number of track and disk.
+const DEFAULT_NUMBER_IN_PAIR: u32 = 1;
+
+// These keys have the part of the number pair.
+const NUMBER_PAIR_KEYS: &[ItemKey] = &[
+	ItemKey::TrackNumber,
+	ItemKey::TrackTotal,
+	ItemKey::DiscNumber,
+	ItemKey::DiscTotal,
+];
 
 macro_rules! impl_accessor {
 	($($name:ident => $id:literal;)+) => {
@@ -320,6 +332,14 @@ impl ID3v2Tag {
 			},
 		};
 	}
+
+	fn insert_number_pair(&mut self, id: &'static str, number: Option<u32>, total: Option<u32>) {
+		if let Some(content) = format_number_pair(number, total) {
+			self.insert(Frame::text(Cow::Borrowed(id), content));
+		} else {
+			log::warn!("{id} is not set. number: {number:?}, total: {total:?}");
+		}
+	}
 }
 
 fn filter_comment_frame_by_description<'a>(
@@ -384,6 +404,21 @@ fn new_picture_frame(picture: Picture, flags: FrameFlags) -> Frame<'static> {
 	}
 }
 
+fn format_number_pair<N, T>(number: Option<N>, total: Option<T>) -> Option<String>
+where
+	N: Display,
+	T: Display,
+{
+	match (number, total) {
+		(Some(number), None) => Some(number.to_string()),
+		(None, Some(total)) => Some(format!(
+			"{DEFAULT_NUMBER_IN_PAIR}{NUMBER_PAIR_SEPARATOR}{total}"
+		)),
+		(Some(number), Some(total)) => Some(format!("{number}{NUMBER_PAIR_SEPARATOR}{total}")),
+		(None, None) => None,
+	}
+}
+
 impl Accessor for ID3v2Tag {
 	impl_accessor!(
 		title  => "TIT2";
@@ -397,14 +432,7 @@ impl Accessor for ID3v2Tag {
 	}
 
 	fn set_track(&mut self, value: u32) {
-		let existing_track_total = self.track_total();
-
-		let trck_content = match existing_track_total {
-			Some(track_total) => format!("{value}/{track_total}"),
-			None => value.to_string(),
-		};
-
-		self.insert(Frame::text(Cow::Borrowed("TRCK"), trck_content));
+		self.insert_number_pair("TRCK", Some(value), self.track_total());
 	}
 
 	fn remove_track(&mut self) {
@@ -416,12 +444,7 @@ impl Accessor for ID3v2Tag {
 	}
 
 	fn set_track_total(&mut self, value: u32) {
-		let track_number = self.track().unwrap_or(1);
-
-		self.insert(Frame::text(
-			Cow::Borrowed("TRCK"),
-			format!("{track_number}/{value}"),
-		));
+		self.insert_number_pair("TRCK", self.track(), Some(value));
 	}
 
 	fn remove_track_total(&mut self) {
@@ -438,14 +461,7 @@ impl Accessor for ID3v2Tag {
 	}
 
 	fn set_disk(&mut self, value: u32) {
-		let existing_disk_total = self.disk_total();
-
-		let tpos_content = match existing_disk_total {
-			Some(disc_total) => format!("{value}/{disc_total}"),
-			None => value.to_string(),
-		};
-
-		self.insert(Frame::text(Cow::Borrowed("TPOS"), tpos_content));
+		self.insert_number_pair("TPOS", Some(value), self.disk_total());
 	}
 
 	fn remove_disk(&mut self) {
@@ -457,12 +473,7 @@ impl Accessor for ID3v2Tag {
 	}
 
 	fn set_disk_total(&mut self, value: u32) {
-		let disk_number = self.disk().unwrap_or(1);
-
-		self.insert(Frame::text(
-			Cow::Borrowed("TPOS"),
-			format!("{disk_number}/{value}"),
-		));
+		self.insert_number_pair("TPOS", self.disk(), Some(value));
 	}
 
 	fn remove_disk_total(&mut self) {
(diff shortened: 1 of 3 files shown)


src/id3/v2/tag.rs
```diff
@@ -285,6 +288,30 @@ impl ID3v2Tag {
 
 		(None, None)
 	}
+
+	fn insert_frame(&mut self, item: TagItem) {
```
uklotzde, line 292: The purpose of this function is to insert an _item_ by mapping it to the underlying representation, but it does not (always) insert a frame -> `insert_item()`

src/id3/v2/tag.rs
```diff
@@ -285,6 +288,30 @@ impl ID3v2Tag {
...
+		fn set_number<F: FnMut(u32)>(item: TagItem, mut setter: F) {
+			let number = item.into_value().text().map(str::parse::<u32>);
+
+			if let Some(Ok(number)) = number {
```
uklotzde, line 296: Log a warning instead of silently ignoring values that couldn't be parsed successfully?

src/id3/v2/tag.rs
```diff
@@ -349,6 +376,14 @@ fn new_picture_frame(picture: Picture, flags: FrameFlags) -> Frame<'static> {
 	}
 }
 
+fn merge_num_pair<N, T>(number: N, total: T) -> String
```
uklotzde, line 379: - `format_number_pair()`? No need for abbreviations here.
- Using optional arguments would eliminate redundant code in setters and make this function more versatile **and** testable in isolation.

src/id3/v2/tag.rs
```diff
@@ -872,10 +915,56 @@ impl<'a> Id3v2TagRef<'a, std::iter::Empty<FrameRef<'a>>> {
...
+		id: &'a str,
+	) -> Option<FrameRef<'a>> {
+		let mut value: Option<Cow<'_, str>> = None;
+
+		if let Some(number) = number {
+			value = Some(Cow::Borrowed(number));
+		}
+
+		if let Some(total) = total {
+			let number = value.unwrap_or_else(|| Cow::Owned(DEFAULT_NUMBER.to_string()));
```
uklotzde, line 930: A `const &str` representation for `DEFAULT_NUMBER` would avoid an unecessary allocation -> `value.as_deref().unwrap_or(DEFAULT_NUMBER_STR)`.

With the proposed changes and extension of `merge_num_pair` this might even become obsolete.

src/id3/v2/tag.rs
```diff
@@ -872,10 +915,56 @@ impl<'a> Id3v2TagRef<'a, std::iter::Empty<FrameRef<'a>>> {
...
+
+			FrameRef {
+				id: frame.id,
+				value: Cow::Owned(frame.value),
+				flags: frame.flags,
+			}
+		})
+	}
+
+	let number_pair_keys = &[
```
uklotzde, line 946: This should be extracted as a constant of the module. Probably near the other _number_ constants. Serves as additional documentation.

src/id3/v2/tag.rs
```diff
@@ -1634,4 +1723,103 @@ mod tests {
 
 		assert_eq!(id3v2, reparsed);
 	}
+
+	#[test]
```
uklotzde, line 1727: The tests only cover a single case. Missing are test for only number and only total.

uklotzde: I am also curious to know what happens if the frame value

- is empty (contains only whitespace)
- "/"
- "/1"
- "1/"

This should be handled by lightweight unit tests, no need for heavy-weight file tests.

uklotzde (changes requested):

mikanbako:
@uklotzde Thank you for review. I will fix this PR.

uklotzde:
> @uklotzde Thank you for review. I will fix this PR.

Take my proposals with a grain of salt. All my remarks in PRs are intended as additional input from an outside viewer who has a different perspective on the topic. They should support you to validate and rethink your design decisions or to reveal blank spaces and edge cases.

A seemingly dumb question is often the most helpful feedback you could get for improving the code.

mikanbako:
I fixed my existing code at first.

I will add tests about the frame values.

@uklotzde Thank you for comment. You have helped me to improve my PR more.

src/id3/v2/tag.rs
```diff
@@ -25,6 +26,8 @@ const COMMENT_FRAME_ID: &str = "COMM";
 const V4_MULTI_VALUE_SEPARATOR: char = '\0';
 const NUMBER_PAIR_SEPARATOR: char = '/';
 
+const DEFAULT_NUMBER: u32 = 1;
```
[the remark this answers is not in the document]
mikanbako: I fixed this part.

src/id3/v2/tag.rs
mikanbako: I fixed this part.

src/id3/v2/tag.rs
mikanbako: I fixed this part.

src/id3/v2/tag.rs
mikanbako: I fixed this part.

src/id3/v2/tag.rs
mikanbako: I did not fixed it. Because `merge_num_pair` was changed.

src/id3/v2/tag.rs
mikanbako: I fixed this part.

src/id3/v2/tag.rs
mikanbako: I added tests about the missing tests at first.

I will add tests about the frame values.

mikanbako:
I fixed a log message only.

I will add tests about the frame values.

src/id3/v2/tag.rs
```diff
@@ -285,6 +297,52 @@ impl ID3v2Tag {
...
+					}
+				}
+			},
+		};
+	}
+
+	fn insert_number_pair(&mut self, id: &'static str, number: Option<u32>, total: Option<u32>) {
+		let content = format_number_pair(number, total);
+
+		match content {
```
uklotzde, line 339: Using `if let Some() = ` would be more idiomatic. I remember that clippy warns when matching on `Option`.

mikanbako:
I added tests about the frame values and fixed the code to except invalid values.

src/id3/v2/tag.rs
mikanbako: Thank you for comment. I fixed the code.

src/id3/v2/tag.rs
```diff
@@ -1634,4 +1734,278 @@ mod tests {
...
+			assert!(tag.disk_total().is_none());
+		}
+
+		assert_invalid("");
+		assert_invalid(" ");
+		assert_invalid("/");
+		assert_invalid("/1");
+		assert_invalid("1/");
+		assert_invalid("a/b");
+		assert_invalid("1 / 2");
```
uklotzde, line 2006: I would expect that leading/trailing whitespaces around the numbers are trimmed.

https://en.wikipedia.org/wiki/Robustness_principle

src/id3/v2/tag.rs
```diff
@@ -1634,4 +1734,278 @@ mod tests {
...
+		assert_invalid("1/2/3");
+		assert_invalid("1//2");
+		assert_invalid(" 1/2 ");
```
uklotzde, line 2009: This input should also be considered as valid.

src/id3/v2/tag.rs
```diff
@@ -547,21 +607,38 @@ impl SplitAndMergeTag for ID3v2Tag {
...
-			if let Some(total) = split.next() {
-				tag.items
-					.push(TagItem::new(total_key, ItemValue::Text(total.to_string())))
-			}
+			let number = split.next()?;
+			let total = split.next();
+
+			let is_valid = if let Some(total) = total {
+				content.len() == number.len() + 1 + total.len()
+					&& str::parse::<u32>(number).is_ok()
```
uklotzde, line 621: By extracting a `parse_number()` function you could handle the whitespace trimming consistently. I consider "total" as a special kind of number, so the naming should be reasonable and we do not need to name it `parse_number_or_total()` .

mikanbako:
@uklotzde Thank you for comments. I will modify my code.

Before I do that, please allow to confirm the details about valid or invalid frames.

I understood the following table, is it right?

<table>
<th>frame value</th>
<th>valid or invalid</th>
<th>number</th>
<th>total
<tr><td>/1</td><td>invalid</td><td>-</td><td>-</td></tr>
<tr><td>1/</td><td>invalid</td><td>-</td><td>-</td></tr>
<tr><td>1&nbsp;/&nbsp;2</td><td>valid</td><td>1</td><td>2</td></tr>
<tr><td>1/2/3</td><td>valid</td><td>1</td><td>2</td></tr>
<tr><td>1//2</td><td>valid</td><td>1</td><td>2</td></tr>
<tr><td>&nbsp;1/2&nbsp;</td><td>valid</td><td>1</td><td>2</td></tr>
<tr><td>a/b&nbsp;</td><td>invalid</td><td>-</td><td>-</td></tr>
<tr><td>1/b&nbsp;</td><td>invalid</td><td>-</td><td>-</td></tr>
<tr><td>a/1&nbsp;</td><td>invalid</td><td>-</td><td>-</td></tr>
</table>

uklotzde:
> Before I do that, please allow to confirm the details about valid or invalid frames.

I suggest to only ignore whitespace, because the overarching object is to preserve all (meaningful) information. Consequently both `1/2/3` and `1//2` should **not** be parsed as a pair of numbers, because we cannot reason about the intention of the user.

I am unsure about `1/` and `/1` which could both be interpreted as a missing numbers (defaulting to 0) when using a sloppy formatter that unconditionally inserts the separator.

mikanbako:
@uklotzde Thank you for the comment. 

I will modify my code to ignore whitespaces.

For the present, `1/` and `/1` will be treated as invalid.

In other words, I plan to modify my code for the following table.

frame value | valid or invalid | number | total
-- | -- | -- | --
/1 | invalid | - | -
1/ | invalid | - | -
1 / 2 | valid | 1 | 2
1/2/3 | invalid | - | -
1//2 | invalid | - | -
&nbsp;1/2&nbsp; | valid | 1 | 2
a/b | invalid | - | -
1/b | invalid | - | -
a/1 | invalid | - | -

mikanbako:
I modified my code to ignore whitespaces while parsing a number pair.

src/id3/v2/tag.rs
```diff
@@ -547,21 +607,50 @@ impl SplitAndMergeTag for ID3v2Tag {
 		fn split_pair(
 			content: &str,
 			tag: &mut Tag,
-			current_key: ItemKey,
+			number_key: ItemKey,
 			total_key: ItemKey,
 		) -> Option<()> {
+			fn parse_number(source: Option<&str>) -> Option<&str> {
```
uklotzde, line 613: This function should accept `source: &str`. Otherwise you are implicitly re-implementing `Option::and_then`.

src/id3/v2/tag.rs
```diff
@@ -547,21 +607,50 @@ impl SplitAndMergeTag for ID3v2Tag {
...
-				.push(TagItem::new(current_key, ItemValue::Text(current)));
 
-			if let Some(total) = split.next() {
-				tag.items
-					.push(TagItem::new(total_key, ItemValue::Text(total.to_string())))
-			}
+			let number_source = split.next()?;
+			let number = parse_number(Some(number_source))?;
+			let total_source = split.next();
+			let total = parse_number(total_source);
```
uklotzde, line 631: `let total = total_source.and_then(parse_number)`

src/id3/v2/tag.rs
```diff
@@ -547,21 +607,50 @@ impl SplitAndMergeTag for ID3v2Tag {
...
+			number_key: ItemKey,
 			total_key: ItemKey,
 		) -> Option<()> {
+			fn parse_number(source: Option<&str>) -> Option<&str> {
+				let number = source?.trim();
+
+				if str::parse::<u32>(number).is_ok() {
+					Some(number)
+				} else {
+					log::warn!("{number:?} could not be parsed as a number.");
```
uklotzde, line 619: Logging a warning if the input is empty is not desired.

mikanbako:
@uklotzde Thank you for the advice. I fixed my code.

src/id3/v2/tag.rs
```diff
@@ -547,21 +607,54 @@ impl SplitAndMergeTag for ID3v2Tag {
 		fn split_pair(
 			content: &str,
 			tag: &mut Tag,
-			current_key: ItemKey,
+			number_key: ItemKey,
 			total_key: ItemKey,
 		) -> Option<()> {
```
uklotzde, line 612: Using `Option<()>` for early returns is sometimes handy but also confusing. Better return nothing and use if-let-else statements for early returns.

src/id3/v2/tag.rs
```diff
@@ -547,21 +607,54 @@ impl SplitAndMergeTag for ID3v2Tag {
...
-			if let Some(total) = split.next() {
-				tag.items
-					.push(TagItem::new(total_key, ItemValue::Text(total.to_string())))
-			}
+			let number_source = split.next()?;
+			let number = parse_number(number_source)?;
+			let total_source = split.next();
+			let total = total_source.and_then(parse_number);
+
+			let is_valid = if let Some(total_source) = total_source {
```
uklotzde, line 637: This check doesn't work. You will notice if your add tests for leading/trailing whitespace.

Implicitly assuming that the length of NUMBER_PAIR_SEPARATOR equals 1 byte is brittle.

src/id3/v2/tag.rs
mikanbako: Please allow me to confirm the details.

`split_pair()` is called in the `match` statement as the [match guard](https://doc.rust-lang.org/book/ch18-03-pattern-syntax.html#extra-conditionals-with-match-guards). `if let` guards are experimental now.

https://github.com/rust-lang/rust/issues/51114

uklotzde: Sorry, I wanted to referr to _let-else statements_ introduced with Rust 1.65. These should be convenient here and allow to avoid the final return of a meaningless result.

src/id3/v2/tag.rs
mikanbako: Please allow me to confirm the details.

> You will notice if you add tests for leading/trailing whitespace.

What test cases do you think? The followings are passed as a valid content when I added the unit test:

* " 1 "
* "&nbsp;&nbsp;1&nbsp;&nbsp;" (The spaces have 2 characters)
* "1&nbsp;&nbsp;/&nbsp;&nbsp;2" (The spaces have 2 characters)
* "&nbsp;&nbsp;1&nbsp;&nbsp;/&nbsp;&nbsp;2&nbsp;&nbsp;" (The spaces have 2 characters)

Because `content` is checked the following:

* A not trimmed string has a number only.
* One separator character and two not trimmed strings that have a number only.

> Implicitly assuming that the length of NUMBER_PAIR_SEPARATOR equals 1 byte is brittle.

May do not NUMBER_PAIR_SEPARATOR equals 1 bytes? I would be glad if you could provide an example.

Otherwise, do you mean that the constant should be extracted?

src/id3/v2/tag.rs
```diff
@@ -547,21 +607,54 @@ impl SplitAndMergeTag for ID3v2Tag {
...
-			let current = split.next()?.to_string();
-			tag.items
-				.push(TagItem::new(current_key, ItemValue::Text(current)));
 
-			if let Some(total) = split.next() {
-				tag.items
-					.push(TagItem::new(total_key, ItemValue::Text(total.to_string())))
-			}
+			let number_source = split.next()?;
+			let number = parse_number(number_source)?;
```
uklotzde, line 633:
```
let Some(number) = parse_number(number_source) else {
    return;
}
```

mikanbako:
@uklotzde Thank you for comments. I replied for them.

src/id3/v2/tag.rs
uklotzde: The magic number number 1 appears out of the context of `NUMBER_PAIR_SEPARATOR`. If the constant was of type `str` you could use `NUMBER_PAIR_SEPARATOR.len()`. But using the size of the separator is probably no longer needed after fixing the parsing. It should not be relevant for the decision, only number and total with their parse results need to be considered.

uklotzde: Returning an `Option` from parse_number() is probably no sufficient, because you need to distinguish between _empty_ and _not a number_.

uklotzde: Using the low-level input representation again for making the final decision is conceptually unsound, different levels of abstraction.

uklotzde: Indeed, I overlooked some tests. But I am still unsure that this code covers all cases correctly, especially the expression `content.len() == number_source.len() + 1 + total_source.len()`. I already mentioned the various reasons why it is not suitable here.

src/id3/v2/tag.rs
```diff
@@ -547,21 +607,54 @@ impl SplitAndMergeTag for ID3v2Tag {
...
 			tag: &mut Tag,
-			current_key: ItemKey,
+			number_key: ItemKey,
 			total_key: ItemKey,
 		) -> Option<()> {
+			fn parse_number(source: &str) -> Option<&str> {
+				let number = source.trim();
+
+				if number.is_empty() {
+					return None;
```
uklotzde, line 617: Returning an empty `str` (= `Some(number)`) would be an option to distinguish between empty and invalid numbers. There are many possibilities for encoding the 3 variants _empty_, _number_, and _invalid_.

src/id3/v2/tag.rs
mikanbako: Thank you for details. I understand that I should improve parsing about `content`. I will consider it.

src/id3/v2/tag.rs
mikanbako: Thank you for details. I will consider to use let-else.

uklotzde:
While trying to get rid of the input string comparisons I came up with a simple and effective solution: https://github.com/mikanbako/lofty-rs/pull/1

mikanbako:
@uklotzde Thank you for PR. It is merged.

uklotzde (approved):
Thank you for this nice cooperation! The code is very solid now after we have considered many possible use cases. The tests will ensure that everything works as expected and are a good starting point if the behavior ever needs to be modified.

@Serial-ATA Please take a final look and then merge. The recent clippy failures seem to be unrelated and should be fixed separately, not by this PR.

Serial-ATA:
Thanks for fixing this! And thanks @uklotzde for the review. This looks good, and has a lot of tests to back it up :smile:.

This makes me realize that the same issue will occur in APE tags as well, which use the same format for track/disk information. I'll have to make an issue for that.

PR #155, opened 2023-02-25
Expose channel mask in file properties

uklotzde:
Currently only supported by WAV files.

- The `ChannelMask` newtype cannot be constructed, i.e. read-only. Edit: Only the `mono()` and `stereo()` constructors are public for convenience and also serve as examples.
- Implementation detail: Explicit destructuring of format-specific file properties upon conversion is safer when new fields are added in the future.

### DIFF
--- a/src/properties.rs
+++ b/src/properties.rs
@@ -1,5 +1,7 @@
 use std::time::Duration;
 
+use crate::ChannelMask;
+
 /// Various *immutable* audio properties
 #[derive(Debug, PartialEq, Eq, Clone)]
 #[non_exhaustive]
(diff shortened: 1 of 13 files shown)


src/properties.rs
```diff
@@ -10,6 +12,7 @@ pub struct FileProperties {
 	pub(crate) sample_rate: Option<u32>,
 	pub(crate) bit_depth: Option<u8>,
 	pub(crate) channels: Option<u8>,
+	pub(crate) channel_mask: Option<ChannelMask>,
```
Serial-ATA, line 15: This'll need a getter like the other fields

Serial-ATA (changes requested):

src/properties.rs
uklotzde: https://github.com/Serial-ATA/lofty-rs/pull/155/commits/6672b3fe8303f269fe12999f04dcf21266e47de7

Serial-ATA:
Thanks! For future reference, this is something that could also easily be supported for mp4a here: https://github.com/Serial-ATA/lofty-rs/blob/31c30389cb89f04086ac856333754177eeead562/src/mp4/properties.rs#L487
With the configurations here: https://wiki.multimedia.cx/index.php?title=MPEG-4_Audio#Channel_Configurations

PR #158, opened 2023-02-28
Fix clippy::let_underscore_untyped

uklotzde:


Serial-ATA:
I don't really understand the point of this lint, it doesn't look much better than `let _`. I think it should just be allowed crate wide: https://github.com/Serial-ATA/lofty-rs/blob/0b627bcbca7e783c3022bfbb46e68cccaf7eba94/src/lib.rs#L115.

Serial-ATA:
Thanks!

Issue #160, opened 2023-03-11
ID3v2: Add support for UFID frames


uklotzde:
Needed for mapping the [MusicBrainz Recording Identifier](https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html#id21).

PR #161, opened 2023-03-11
Add support for selected MusicBrainz identifiers

uklotzde:
- Only the single-valued recording/track/album/release identifiers, no artist identifiers.
- No mapping of `MusicBrainzRecordingId` for ID3v2 yet, see #160.


src/tag/item.rs
```diff
@@ -208,7 +212,10 @@ gen_map!(
 	"TKWD"                  => PodcastKeywords,
 	"COMM"                  => Comment,
 	"TLAN"                  => Language,
-	"USLT"                  => Lyrics
+	"USLT"                  => Lyrics,
+	"MusicBrainz Release Track Id" => MusicBrainzTrackId,
+	"MusicBrainz Album Id"  => MusicBrainzReleaseId,
+	"MusicBrainz Release Group Id" => MusicBrainzReleaseGroupId
```
Serial-ATA, line 218: Could you fixup the alignment of the arrows here?

src/tag/item.rs
```diff
@@ -278,7 +285,11 @@ gen_map!(
 	"----:com.apple.iTunes:LANGUAGE"              => Language,
 	"----:com.apple.iTunes:SCRIPT"                => Script,
 	"\u{a9}lyr"                                   => Lyrics,
-	"xid "                                        => AppleXid
+	"xid "                                        => AppleXid,
+	"----:com.apple.iTunes:MusicBrainz Track Id"  => MusicBrainzRecordingId,
+	"----:com.apple.iTunes:MusicBrainz Release Track Id" => MusicBrainzTrackId,
+	"----:com.apple.iTunes:MusicBrainz Album Id"  => MusicBrainzReleaseId,
+	"----:com.apple.iTunes:MusicBrainz Release Group Id" => MusicBrainzReleaseGroupId
```
Serial-ATA, line 292: Same as above

Serial-ATA (changes requested):

src/tag/item.rs
uklotzde: That's why I prefer **not** to use any manual alignment. It causes changes in totally unrelated and unchanged lines.

Serial-ATA:
Thanks!

PR #173, opened 2023-03-22
picture: add `into_data()`

hinto-janai:
This adds some functions that exposes the inner data.

Since the inner `Cow` is `Cow::Owned`, `into_owned()` _should_ be a cheap no-clone/no-allocation call.

Also, most libraries seem use `bytes()` or `as_bytes()` instead of `data()` but that'll break everyone's code, so, maybe not worth it.

### DIFF
--- a/src/picture.rs
+++ b/src/picture.rs
@@ -535,9 +535,14 @@ impl Picture {
 		self.description = description.map(Cow::from);
 	}
 
-	/// Returns the picture data
+	/// Returns the [`Picture`] data as borrowed bytes.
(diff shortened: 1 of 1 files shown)


Serial-ATA (changes requested):
Thanks! The reason I chose `data()` instead of `as_bytes()` is `Picture` represents more than just the image file. That's why there are `as_{ape, apic, flac}_bytes()` methods.

hinto-janai:
Ah, I see.

Requested changes were made.

Serial-ATA:
Thanks!

PR #180, opened 2023-04-11
id3v2: Fix parsing of multi-valued text fields

uklotzde:
Otherwise empty items are added. The regression test demonstrates this.


uklotzde:
This occured for some files with multiple `0` terminators in their. But the current behavior is actually correct.

PR #205, opened 2023-06-03
probe: Add new default ParsingMode variant

### LINKED ISSUE #204: ID3v2: Missing owner in UFID frame
Hi, 
firstly thanks for the easy-to-use library, it's been a breeze so far!
[Symphonia](https://github.com/pdeljanov/Symphonia) is just too vast for my use case.

I'm trying to read the metadata from a music library with all sorts of file types, I find all working pretty well out of the box.

I encountered this error `ERROR: Failed to read file!: Id3v2(ID3v2: MissingUfidOwner)` while reading from a `.mp3`.
The file is not currupted, plays well and when reading tags with symphonia I get the following:

```
|     [01] Codec:           MPEG Audio Layer 3 (mp3)
|          Sample Rate:     44100
|          Duration:        0:06:40.039 (17641728)
|          Time Base:       1/44100
|          Encoder Delay:   1105
|          Encoder Padding: 623
|          Channel(s):      2
|          Channel Map:     0b000000000000000000000000000011
```

The part which is failing me:
```
let tagged_file = Probe::open(path)
	.expect("ERROR: Bad path provided!")
	.read()
	.expect("ERROR: Failed to read file!"); // <---- 
```

When browsing the documentation, I found a piece about [registering a custom resolver](https://github.com/Serial-ATA/lofty-rs/blob/main/examples/custom_resolver/src/main.rs), but don't really know how to go about it in this case, as I dont have a custom type at hand.

At this point I dont really know what to try next, as I feel like it's a really bad idea to create the `TaggedFile` myself.

Any hints or things to try?
Happy to share the audio file (copyrighted...) via zippy, trying first without.

PS: I saw a call for more documentation, I'm happy to create digest of whatever comes out of this for future reference.

Serial-ATA:
The new `BestAttempt` variant will be less eager to error than the previous default, `ParsingMode::Strict`. This comes with the consequence that the input may not be *entirely* spec-compliant.

This implements `BestAttempt` for `UniqueFileIdentifierFrame::parse` and adds `Strict` checks for `VorbisComments` parsing. More checks will come in the future.

closes #204

### DIFF
--- a/src/probe.rs
+++ b/src/probe.rs
@@ -115,6 +115,20 @@ impl ParseOptions {
 }
 
 /// The parsing strictness mode
+///
+/// This can be set with [`Probe::options`].
+///
+/// # Examples
+///
+/// ```rust,no_run
+/// use lofty::{ParseOptions, ParsingMode, Probe};
+///
+/// # fn main() -> lofty::Result<()> {
+/// // We only want to read spec-compliant inputs
+/// let parsing_options = ParseOptions::new().parsing_mode(ParsingMode::Strict);
+/// let tagged_file = Probe::open("foo.mp3")?.options(parsing_options).read()?;
+/// # Ok(()) }
+/// ```
 #[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq, Default)]
 #[non_exhaustive]
 pub enum ParsingMode {
--- a/src/probe.rs
+++ b/src/probe.rs
@@ -20,7 +20,7 @@ use std::io::{BufReader, Cursor, Read, Seek, SeekFrom};
 use std::path::Path;
 
 /// Options to control how Lofty parses a file
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq)]
 #[non_exhaustive]
 pub struct ParseOptions {
 	pub(crate) read_properties: bool,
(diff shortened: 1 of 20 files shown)


src/probe.rs
```diff
@@ -115,12 +115,25 @@ impl ParseOptions {
...
+	/// This mode will attempt to fill in any holes where possible in otherwise valid, spec-compliant input.
+	///
+	/// NOTE: A readable input does *not* necessarily make it writeable.
+	#[default]
+	BestAttempt,
+	/// Least eager to error, may produce invalid output
+	///
+	/// This mode will discard any invalid fields, and ignore the majority of non-fatal errors.
+	///
+	/// Tags and properties read using this mode may come out missing items if the input is malformed.
```
uklotzde, line 150: The wording sounds complicated to me as a non-natice speaker. One might overlook that the final result could be incomplete if errors occurred.  It would help to explicitly use the terms "imcomplete" and/or "partial" in the wording.

src/probe.rs
```diff
@@ -115,12 +115,25 @@ impl ParseOptions {
...
+	/// Will eagerly error on invalid input
+	///
+	/// This mode will eagerly error on any non spec-compliant input.
 	Strict,
-	/// Less eager to error, may produce invalid output
+	/// Default mode, less eager to error on recoverably malformed input
+	///
+	/// This mode will attempt to fill in any holes where possible in otherwise valid, spec-compliant input.
+	///
+	/// NOTE: A readable input does *not* necessarily make it writeable.
```
uklotzde, line 143: Concise disclaimer :+1:  That's exactly what clients need to be prepared for.

src/probe.rs
Serial-ATA: Good call :+1:

PR #207, opened 2023-06-04
misc: Add allocation limit to `try_vec!`

Serial-ATA:
This adds an arbitrary allocation limit of 8MB for any one item, which should be more than enough in any case.

This was changed to stop the fuzz tests from failing with OOM errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 - **Probe**: The default `ParsingMode` is now `ParsingMode::BestAttempt` (It was previously `ParsingMode::Strict`)
+- **Alloc**:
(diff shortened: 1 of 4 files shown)


hinto-janai:
Does this include `Picture`'s?

I'm sure I have some audio files with enormous embedded images (although probably not >8MB...?)

I'll scan all of them and report back.

Serial-ATA:
Yeah, this includes pictures as well. I haven't encountered any image >8MB in my experience.

hinto-janai:
I have a few that are `5.1 MB`. Is there a downside to increasing the limit to just before the fuzz tests start failing?

Serial-ATA:
The fuzz tests would OOM at different points depending on the system.

I think 8MB is probably a safe size. Having a single tag item >8MB seems a little crazy to me. Unless there's something I'm not thinking of.

hinto-janai:
Ok, sounds good.

Serial-ATA:
Actually, it wouldn't hurt to make this configurable in `ParseOptions` eventually. I'll make an issue for that.

PR #212, opened 2023-07-02
ID3v2: Report bad or unsupported frame IDs

uklotzde:
This is the first step to reveal information about malformed/unsupported tags.

Those frames should be skipped when parsing in `Relaxed` mode.

Note: The `BadFrameId` error was used for different purposes, not only for parsing as the message suggests.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -4,6 +4,7 @@
 //! which can be extended at any time.
 
 use crate::file::FileType;
+use crate::ItemKey;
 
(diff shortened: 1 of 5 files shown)


Serial-ATA:
Thanks, I reverted the purely fmt changes. I don't think it looks too good, I'll have to configure rustfmt to cut that out.

uklotzde:
I had to update my (barely) used nightly toolchain just to fix the CI errors. This is inconvenient.

PR #214, opened 2023-07-02
ID3v2: Ignore frames with bad IDs while reading

uklotzde:
Follow-up of #212.


uklotzde:
CI fails again due to unrelated rustfmt errors.

src/id3/v2/frame/header.rs
```diff
@@ -24,12 +26,20 @@ where
...
+		.map_err(|_| Id3v2Error::new(Id3v2ErrorKind::BadFrameId(frame_id_bytes.to_vec())).into())
+		.map(|id_str| {
+			upgrade_v2(id_str).map_or_else(|| Cow::Owned(id_str.to_owned()), Cow::Borrowed)
+		})
+		.and_then(FrameId::new_cow)
+	{
+		Ok(id) => id,
+		Err(err) => match parse_mode {
+			ParsingMode::Strict => return Err(err),
+			ParsingMode::BestAttempt | ParsingMode::Relaxed => return Ok(None),
```
Serial-ATA, line 40: `Ok(None)` is currently used to indicate that we encountered padding or the reader is exhausted. This won't only ignore the bad frame ID, but stop reading the tag entirely.

I think errors should continue to be thrown from `parse_*_header`, and handled once in `Frame::read`. That'd simplify things a bit, since the match on `parse_mode` appears 3 times in this file.

From there, you can return `Ok((None, false))` from `Frame::read` to indicate that there *are* more frames to come after the failure:
https://github.com/Serial-ATA/lofty-rs/blob/f1e8ab6c2fb1bb275a15d9e4246ad49f2dbbb3da/src/id3/v2/read.rs#L50-L56

uklotzde: The `Option` is the wrong choice here and very confusing. It needs to be replaced by a dedicated type that reveals the semantics.

src/id3/v2/frame/read.rs
```diff
@@ -11,23 +11,43 @@ use std::io::Read;
...
+			Ok(None) => {
+				// Stop reading
+				return Ok(Self::Eof);
+			},
+			Ok(Some(some)) => some,
+			Err(err) => {
+				match parse_mode {
+					ParsingMode::Strict => return Err(err),
+					ParsingMode::BestAttempt | ParsingMode::Relaxed => {
+						// Skip this frame and continue reading
```
Serial-ATA, line 45: I added a test for this, and just realized that the frame will not actually be skipped over.

There would need to be a way to get the `size` out of the header even if there's an error.

uklotzde: Fixed.

Code could be improved later to avoid the mutable reference param. Not urgent.

uklotzde: The naming of `ParsedFrame` is more appropriate now, because the skipping actually happens later.

Serial-ATA:
Thanks!

I'll move Clippy/rustfmt to stable, the failures are annoying.

PR #215, opened 2023-07-02
Support 'Involved People List' Id3v2 frame

### LINKED ISSUE #213: ID3v2 IPLS/TPIL Involved People List is not read correctly
### Reproducer

Run the tag_reader example with the attached file (Audio has been removed for copyright reasons but the tag remains, also rename it to mp3, I renamed it to txt since Github doesn't allow mp3 upload)

### Summary

I stepped through the lofty code with a debugger and the TIPL tag (IPLS in the file) is read incorrectly. IPLS is a list of terminated strings ([source](https://web.archive.org/web/20150222220706/http://id3.org/id3v2.3.0#Involved_people_list)). Lofty tries to read it as one string and so only the first item is read. Then lofty will try to read the next frame in the data of the IPLS tag and crash.

### Expected behavior

The IPLS frame is read correctly

### Assets

!!! RENAME TO MP3 !!!
[testz.txt](https://github.com/Serial-ATA/lofty-rs/files/11929284/testz.txt)

FriederHannenheim:
fixes #213


uklotzde:
I suggest to add a round trip test (write -> read -> verify) an ID3v2 tag that contains a `KeyValueFrame`.

src/id3/v2/frame/content.rs
```diff
@@ -28,6 +29,7 @@ pub(super) fn parse_content<R: Read>(
 		"WXXX" => ExtendedUrlFrame::parse(reader, version)?.map(FrameValue::UserUrl),
 		"COMM" => CommentFrame::parse(reader, version)?.map(FrameValue::Comment),
 		"USLT" => UnsynchronizedTextFrame::parse(reader, version)?.map(FrameValue::UnsynchronizedText),
+		"TIPL" => KeyValueFrame::parse(reader, version)?.map(FrameValue::KeyValueFrame),
```
Serial-ATA, line 32: I believe the same can be done for TMCL frames as well

src/id3/v2/items/key_value_frame.rs
```diff
@@ -0,0 +1,71 @@
...
+	/// Convert a [`KeyValueFrame`] to a byte vec
+	pub fn as_bytes(&self) -> Vec<u8> {
+		let mut content = vec![];
+
+		for (key, value) in &self.key_value_pairs {
+			content.append(&mut encode_text(key, self.encoding, true));
+			content.append(&mut encode_text(value, self.encoding, true));
+		}
+
+		content.insert(0, self.encoding as u8);
```
Serial-ATA, line 68:
```suggestion
		let mut content = vec![self.encoding as u8];

		for (key, value) in &self.key_value_pairs {
			content.append(&mut encode_text(key, self.encoding, true));
			content.append(&mut encode_text(value, self.encoding, true));
		}
```

src/id3/v2/items/key_value_frame.rs
```diff
@@ -0,0 +1,71 @@
...
+	where
+		R: Read,
+	{
+		let Ok(encoding_byte) = reader.read_u8() else {
+			return Ok(None);
+		};
+
+		let encoding = verify_encoding(encoding_byte, version)?;
+
+		let mut values = vec![];
```
Serial-ATA, line 41:
```suggestion
		let mut values = Vec::new();
```
This is just a nitpick, but I prefer `Vec::new()` over empty `vec!` calls.

src/id3/v2/frame/mod.rs
```diff
@@ -173,6 +174,8 @@ pub enum FrameValue {
 	Picture(AttachedPictureFrame),
 	/// Represents a "POPM" frame
 	Popularimeter(Popularimeter),
+	/// Represents an "IPLS" or "TPIL" frame
+	KeyValueFrame(KeyValueFrame),
```
Serial-ATA, line 178:
```suggestion
	KeyValue(KeyValueFrame),
```
This should follow the other variants' definitions: `X(XFrame)`

src/id3/v2/write/frame.rs
```diff
@@ -47,6 +47,7 @@ fn verify_frame(frame: &FrameRef<'_>) -> Result<()> {
 				FrameValue::UserUrl(_) => "UserUrl",
 				FrameValue::Picture { .. } => "Picture",
 				FrameValue::Popularimeter(_) => "Popularimeter",
+				FrameValue::KeyValueFrame(_) => "KeyValueData",
```
Serial-ATA, line 50:
```suggestion
				FrameValue::KeyValue(_) => "KeyValue",
```

Serial-ATA (changes requested):

FriederHannenheim:
Thanks for your suggestions. I implemented the requested changes and tried to create a test, however the test is failing. I don't know why. Could one of you maybe take a look?

tests/files/mpeg.rs
```diff
@@ -306,3 +307,57 @@ fn remove_id3v1() {
...
 }
+
+#[test]
+fn read_and_write_tpil_frame() {
+	let key_value_pairs = vec![
+					("engineer".to_string(), "testperson".to_string()), 
+					("vocalist".to_string(), "testhuman".to_string())
+				];
+
+	let mut file = temp_file!("tests/files/assets/issue_213.mp3");
```
Serial-ATA, line 318: This file doesn't exist, did you forget to commit it?

tests/files/mpeg.rs
```diff
@@ -306,3 +307,57 @@ fn remove_id3v1() {
...
+
+	let tagged_file = Probe::new(&mut file)
```
Serial-ATA, line 320: Since this isn't a probe test, I'd just use `MpegFile::read_from` so that you can just use `Id3v2Tag` directly.

tests/files/mpeg.rs
```diff
@@ -306,3 +307,57 @@ fn remove_id3v1() {
...
+	file.rewind().unwrap();
+	let tagged_file = Probe::new(&mut file)
+		.options(ParseOptions::new().read_properties(false))
+		.guess_file_type()
+		.unwrap()
+		.read()
+		.unwrap();
+	
+
+	let tag: Id3v2Tag = tagged_file.tag(TagType::Id3v2).unwrap().clone().into();
```
Serial-ATA, line 355: The TIPL frame won't exist since you're reading a `TaggedFile`. The conversion from `Id3v2Tag` -> `Tag` explicitly discards it here:
```rs
						FrameValue::KeyValue(_) | FrameValue::UniqueFileIdentifier(_) => {
							return true; // Keep unsupported frame
						},
```
You'd have to read an `MpegFile` to use the unaltered `Id3v2Tag` directly.

It may (sometime in the future) be worth taking values from the TIPL to convert to generic `ItemKey`s, such as `TIPL:arranger` mapping to `ItemKey::Arranger`, etc.

Serial-ATA (changes requested):

tests/files/mpeg.rs
uklotzde: Please don't add more binary files to the repo!

This is not needed for these kind of round-trip tests. It is easier to generate the expected contents of the file programmatically. That's why I proposed to first write and then (re-)read the written data.

uklotzde: Here you find some examples: https://github.com/Serial-ATA/lofty-rs/blob/f1e8ab6c2fb1bb275a15d9e4246ad49f2dbbb3da/tests/files/mpeg.rs#L250

tests/files/mpeg.rs
uklotzde: The test should verify that the incomplete/asymmetric implementation should not do any harm to existing files for the read-modify-write use case.

FriederHannenheim:
I've fixed the test

Serial-ATA:
Looks good, thanks!

PR #216, opened 2023-07-02
APE: Properly convert track/disk number pairs when writing

### LINKED ISSUE #159: APE track/disk numbers are not properly converted when writing
### Reproducer

I tried this code:

```rust
use lofty::{Accessor, Tag, TagExt, TagType};

fn main() {
    let mut tag = Tag::new(TagType::APE);
    tag.set_track(5);
    tag.set_track_total(15);
    tag.save_to_path("foo.mp3").unwrap();
}
```

### Summary

When using `Tag`, the track/disk numbers are not properly converted when saving to an APE tag. This is identical to #145, as it will be saved as two separate items: `........Track.5........Track.15`.

This issue was fixed for ID3v2 in #149, which stores number pairs in the same format. The fix should be similar, if not identical.

### Expected behavior

_No response_

### Assets

_No response_

Serial-ATA:
This also changes the default track/disk number to `DEFAULT_NUMBER_IN_PAIR`.

closes #159

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ byteorder  = "1.4.3"
 # ID3 compressed frames
 flate2     = { version = "1.0.26", optional = true }
 # Proc macros
-lofty_attr = "0.8.0"
+lofty_attr = { path = "lofty_attr" }
 # Debug logging
 log        = "0.4.18"
 # OGG Vorbis/Opus
--- a/src/id3/v2/util/pairs.rs
+++ b/src/id3/v2/util/pairs.rs
@@ -1,4 +1,5 @@
-/// Contains utilities for ID3v2 style number pairs
+//! Contains utilities for ID3v2 style number pairs
+
 use crate::tag::item::{ItemKey, TagItem};
 
 use std::fmt::Display;
--- a/src/id3/v2/util/pairs.rs
+++ b/src/id3/v2/util/pairs.rs
@@ -35,18 +35,15 @@ where
 
 /// Attempts to convert a `TagItem` to a number, passing it to `setter`
 pub(crate) fn set_number<F: FnMut(u32)>(item: &TagItem, mut setter: F) {
-	let text = item.value().text().map(str::trim);
+	let text = item.value().text();
 
-	let trimmed;
-	match text {
-		None | Some("") => {
-			log::warn!("Value does not have text in {:?}", item.key());
-			return;
-		},
-		Some(trimmed_text) => trimmed = trimmed_text,
+	let trimmed_text = text.unwrap_or_default().trim();
+	if trimmed_text.is_empty() {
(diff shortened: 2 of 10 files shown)


src/id3/v2/util/pairs.rs
```diff
@@ -0,0 +1,54 @@
...
+		)),
+		(Some(number), Some(total)) => Some(format!("{number}{NUMBER_PAIR_SEPARATOR}{total}")),
+		(None, None) => None,
+	}
+}
+
+/// Attempts to convert a `TagItem` to a number, passing it to `setter`
+pub(crate) fn set_number<F: FnMut(u32)>(item: &TagItem, mut setter: F) {
+	let text = item.value().text();
+	let number = text.map(str::parse::<u32>);
```
uklotzde, line 39: What happens if a field is empty or solely contains whitespace characters? These cases should probably be interpreted like a missing field.

The parsing should also not fail if the number is enclosed in leading/trailing whitespace characters, e.g. parsing  `" 12 "` should not fail.

uklotzde (changes requested):
Parsing of numbers needs to be more robust and more liberal. The special cases should be tested.

src/id3/v2/util/pairs.rs
uklotzde: The text should be trimmed and then checked for emptiness before trying to parse the number.

Cargo.toml
```diff
@@ -18,7 +18,7 @@ byteorder  = "1.4.3"
 # ID3 compressed frames
 flate2     = { version = "1.0.26", optional = true }
 # Proc macros
-lofty_attr = "0.8.0"
+lofty_attr = { path = "lofty_attr" }
```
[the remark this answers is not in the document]
Serial-ATA: Yes, it used `ape::tag::tagitems_into_ape`, which was changed.

src/id3/v2/util/pairs.rs
```diff
@@ -0,0 +1,87 @@
...
+			"{DEFAULT_NUMBER_IN_PAIR}{NUMBER_PAIR_SEPARATOR}{total}"
+		)),
+		(Some(number), Some(total)) => Some(format!("{number}{NUMBER_PAIR_SEPARATOR}{total}")),
+		(None, None) => None,
+	}
+}
+
+/// Attempts to convert a `TagItem` to a number, passing it to `setter`
+pub(crate) fn set_number<F: FnMut(u32)>(item: &TagItem, mut setter: F) {
+	let text = item.value().text().map(str::trim);
```
uklotzde, line 38: Too many local variables with confusing naming. How about this:

```
let trimmed_text = item.value().text().unwrap_or_default().trim();
if trimmed_text.is_empty() {
    ...early return...
}
```

PR #227, opened 2023-07-05
probe: Add `ParseOptions::max_junk_bytes`

Serial-ATA:
This allows us to sift through junk bytes to find what we need.

Discovered through discussion #219.


src/file.rs
```diff
@@ -1008,3 +1011,18 @@ impl FileType {
...
+/// The result of a `FileType` guess
+///
+/// External callers of `FileType::from_buffer()` will only ever see `Determined` cases.
+/// The remaining cases are used internally in `Probe::guess_file_type()`.
+pub(crate) enum FileTypeGuessResult {
+	/// The `FileType` was guessed
+	Determined(FileType),
+	/// The stream starts with an ID3v2 tag
+	MaybePrecededById3(u32),
+	/// The stream starts with junk zero bytes
```
uklotzde, line 1024: Why are the zero bytes at the beginning called *junk* while at the end they are denoted as *padding*? Would it be suitable to consider both as *padding* to avoid introducing a new terminology?

uklotzde: Or might *junk* also contain non-zero bytes? Then this comment would be imprecise.

Serial-ATA: It would be junk in either case. Once it's out of the context of a tag, anything that was once padding is now garbage data. Where is it referred to as padding? It should be updated.

> Or might junk also contain non-zero bytes? Then this comment would be imprecise.

It's currently only zeroes, but it could be changed to be anything. It's only a matter of time before some files come along with some other types of junk, as seen in the case of #28.

sandersantema:
This fixes the issue for me for all the tracks I had which weren't working, so afaik it is fixed :tada: Thanks a lot.

sandersantema:
Just to confirm: this is expected to work without changing any options or setting the `ParseOptions` correct? That's the behavior I'm seeing in any case.

Serial-ATA:
Yep, `ParseOptions::max_junk_bytes` defaults to 1024.

Issue #189, opened 2023-04-16
Support all special ID3v2 frames


Serial-ATA:
## Native frames

- [ ] [AENC](https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html#aenc) Audio encryption
- [X] [APIC](https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html#apic) Attached picture
- [ ] [ASPI](https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html#aspi) Audio seek point index
- [X] [COMM](https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html#comm) Comments
- [ ] [COMR](https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html#comr) Commercial frame
- [ ] [ENCR](https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html#encr) Encryption method registration
- [ ] [EQU2](https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html#equ2) Equalisation (2)
- [ ] [ETCO](https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html#etco) Event timing codes
- [X] [GEOB](https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html#geob) General encapsulated object
- [ ] [GRID](https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html#grid) Group identification registration
- [ ] [LINK](https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html#link) Linked information
- [X] [MCDI](https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html#mcdi) Music CD identifier (Just a binary frame, no extra support needed)
- [ ] [MLLT](https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html#mllt) MPEG location lookup table
- [ ] [OWNE](https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html#owne) Ownership frame
- [ ] [PRIV](https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html#priv) Private frame
[...]

PR #236, opened 2023-07-18
ID3v2: Support more frames

Serial-ATA:
This adds support for the following frames: "RVA2", "OWNE", "ETCO", and "PRIV".

Additionally, this changes `Id3v2Tag::remove` to return the frame(s) removed, and `Id3v2Tag::insert` to check for frames that only meant to appear once in a file, replacing if necessary. Those frames are: "MCDI", "ETCO", "MLLT", "SYTC", "RVRB", "PCNT", "RBUF", "POSS", "OWNE", "SEEK", and "ASPI".

Part of #189


src/id3/v2/frame/mod.rs
```diff
@@ -181,6 +181,16 @@ pub enum FrameValue {
 	Popularimeter(Popularimeter),
 	/// Represents an "IPLS" or "TPIL" frame
 	KeyValue(KeyValueFrame),
+	/// Represents an "RVA2" frame
+	RelativeVolumeAdjustment(RelativeVolumeAdjustmentFrame),
+	/// Unique file identifier
+	UniqueFileIdentifier(UniqueFileIdentifierFrame),
```
uklotzde, line 187: Great. This is used by many MusicBrainz tags.

src/id3/v2/frame/mod.rs
```diff
@@ -289,10 +321,35 @@ impl FrameValue {
...
 			FrameValue::UniqueFileIdentifier(frame) => frame.as_bytes(),
+			FrameValue::Ownership(frame) => frame.as_bytes()?,
+			FrameValue::EventTimingCodes(frame) => frame.as_bytes(),
+			FrameValue::Private(frame) => frame.as_bytes(),
+			FrameValue::Binary(binary) => binary.clone(),
 		})
 	}
+
+	/// Used for errors in write::frame::verify_frame
+	pub(super) fn name(&self) -> &'static str {
```
uklotzde, line 334: I use `strum` for this purpose: https://docs.rs/strum/latest/strum/trait.VariantNames.html

src/id3/v2/tag.rs
```diff
@@ -320,8 +331,18 @@ impl Id3v2Tag {
 	}
 
 	/// Removes a [`Frame`] by id
-	pub fn remove(&mut self, id: &str) {
-		self.frames.retain(|f| f.id_str() != id)
+	pub fn remove(&mut self, id: &str) -> impl Iterator<Item = Frame<'static>> + '_ {
```
uklotzde, line 334: Passing a `FrameId<'_>` would be type safe and avoid the somehow unrelated `eq_ignore_ascii_case()`. Otherwise we should extract a function that explcitly matches `FrameId` with `str` consistently instead of inlining this code here and probably elsewhere.

src/id3/v2/frame/mod.rs
Serial-ATA: Oh yeah, looks like they have some useful derives. There are probably a few places they could be used. I might look into it eventually.

src/id3/v2/tag.rs
Serial-ATA: Yeah, I'll change that soon in a follow up PR. I don't know how many places it's used right now.

Issue #260, opened 2023-10-01
Probe::guess_file_type() fails for some MP3 files


uklotzde:
### Reproducer

Fails: `Probe::new(BufReader::new(File::open("/tmp/test.mp3").unwrap())).guess_file_type()`
Succeeds: `Probe::open("/tmp/test.mp3")`

### Summary

`Probe::guess_file_type()` fails for some MP3 files.

### Expected behavior

If opening (and reading) a file path directly works then `Probe::guess_file_type()` must not fail.

uklotzde:
Workaround: `.options(ParseOptions::new().max_junk_bytes(usize::MAX))`

Serial-ATA:
So the default junk search window is 1024 bytes, which in most cases should be enough.

Is the data at the start of your files junk? And if so, does it really occupy so much space that 1024 bytes isn't big enough?

`Probe::open` is going to simply look at the `.mp3` extension to guess the file type and pass it directly to `MpegFile::read_from`, which is more lenient in that it has no maximum for junk bytes allowed. `Probe::guess_file_type` will only be able to search what you allow it to.

If the data is junk, there's not anything that can really be done here.

uklotzde:
I have found a bunch of files that are affected. The inconsistent behavior may affect and confuse users. Not unlikely that those files appear in the wild more often.

Unfortunately, I don't have a solution other than using a custom, unlimited look-ahead as a workaround. I need file type specific pre/post-actions during the import and have no choice other than using `guess_file_type()` for the dispatch.

uklotzde:
Possible solution: Unconditionally use `max_junk_bytes = usize::MAX` for known and supported file extensions to reduce the chance for false negatives.

Serial-ATA:
The problem with increasing the junk window is that it becomes much more expensive for users who are simply reading all files in a directory, for example. Now, more time will be spent trying to determine the formats of non-audio/unsupported formats in the directory, as it will exhaust the window, copying all of the bytes, until it ultimately determines it isn't valid.

Having multiple files that are so heavily packed with junk seems odd. Do you know if the data is actually junk or is it something like a tag that just isnt supported?

uklotzde:
> Do you know if the data is actually junk or is it something like a tag that just isnt supported?

I could provide affected files for further analysis privately.

Serial-ATA:
Sure, you can send it to `serial AT [domain on my profile]`. Note there's a 25MB attachment limit.

[...]

PR #261, opened 2023-10-02
mp4: Skip unexpected or empty "data" ilst atoms

uklotzde:
Make use of `ParsingMode` and ignore invalid/unexpected `ilst` atoms that could safely be ignored.

I found a few affected files that are accepted by other MP4 parsers.

PS: It would be cool to include this in a v0.16.1 release soon.

### DIFF
--- a/src/mp4/ilst/read.rs
+++ b/src/mp4/ilst/read.rs
@@ -10,11 +10,16 @@ use crate::mp4::ilst::atom::AtomDataStorage;
 use crate::mp4::read::{skip_unneeded, AtomReader};
 use crate::picture::{MimeType, Picture, PictureType};
 use crate::util::text::utf16_decode;
+use crate::ParsingMode;
 
 use std::borrow::Cow;
 use std::io::{Cursor, Read, Seek, SeekFrom};
 
-pub(in crate::mp4) fn parse_ilst<R>(reader: &mut AtomReader<R>, len: u64) -> Result<Ilst>
+pub(in crate::mp4) fn parse_ilst<R>(
+	reader: &mut AtomReader<R>,
+	parsing_mode: ParsingMode,
+	len: u64,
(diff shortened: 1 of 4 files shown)


src/mp4/ilst/read.rs
```diff
@@ -187,17 +202,32 @@ where
 		// We don't care about the locale
 		reader.seek(SeekFrom::Current(4))?;
 
-		let content_len = (data_atom.len - 16) as usize;
-		if content_len == 0 {
-			// We won't add empty atoms
-			return Ok(None);
```
uklotzde, line 193: This early return was wrong. The comment states that multiple "data" atoms could occur, but the function aborts on the first empty atom.

Serial-ATA:
Thanks! I'm not surprised atoms other than `data` can appear. Bunch of weird unspecified features of MP4 files with everyone doing their own thing.

I'll get 0.16.1 out after looking into #260.

uklotzde:
@Serial-ATA How about releasing v0.16.1 after various fixes have been merged?

Serial-ATA:
I was going to release 0.16.1 after looking into that UTF-16 multi-value frame issue that was a part of #260. I just haven't been able to come up with a good solution yet.

uklotzde:
> I was going to release 0.16.1 after looking into that UTF-16 multi-value frame issue that was a part of #260. I just haven't been able to come up with a good solution yet.

How do you extract the sample test data from the files?

Creating a draft PR with a failing test could be a good starting point for anyone to pick this up.

Serial-ATA:
I just use a hex editor to either create tags by hand or just strip them out of files. 

It may be beneficial to make a simple binary to dump out all of the tags. That would save people from having to do that or email me their assets. Can't say when/if I'll get that done, though.

PR #263, opened 2023-10-03
Replace base64 with data-encoding

uklotzde:
Just stumbled over this discussion: https://github.com/marshallpierce/rust-base64/issues/213

There is also `base64-simd`, but I don't trust it (yet).


uklotzde:
Not urgent. I'll keep the branch, just in case.

PR #275, opened 2023-10-22
ID3v2: Trim null terminators when reading some frames

### LINKED ISSUE #273: ID3v2: Handle null terminators in UTF-8 encoded frames.
iTunes and applications that use Mutagen will add null terminators to UTF-8 encoded frames due to ambiguities in the spec. See this issue for some background:

https://github.com/quodlibet/mutagen/issues/379

This causes lofty to create an extra empty value when splitting multiple values.

Given how widespread a practise this is, it seems necessary to handle it when reading the content of a frame. I'm not sure exactly which frames can have null terminators, but I've noticed it in Text, ExtendedText, and Comment frames.

Here's an example file:

https://archive.org/download/The_Four_Seasons_Vivaldi-10361/John_Harrison_with_the_Wichita_State_University_Chamber_Players_-_05_-_Summer_Mvt_2_Adagio.mp3

sublipri:
Various applications will add null terminators to some frames with UTF-8 and Latin1 encoding. Multiple trailing nulls have also been observed in some files in the wild.  This pull request adds a helper function and uses it to trim any trailing nulls on Comment, Text, UserText, UserUrl, and UnsynchronizedText frames.

closes #273


src/util/text.rs
```diff
@@ -212,6 +212,13 @@ pub(crate) fn encode_text(text: &str, text_encoding: TextEncoding, terminated: b
 	}
 }
 
+pub(crate) fn trim_end_nulls(text: &mut String) {
```
uklotzde, line 215: Why `trim_end...` vs. `trim_trailing_...`? I would prefer a consistent naming of these functions.

sublipri: I chose `trim_end_nulls` to be consistent with `trim_end_matches` from `std::str`. I guess I used `trim_trailing_...` in the test name because I thought it was a bit more descriptive. Should I change the test name to `trim_end...`?

uklotzde: I didn't notice that this only affects the test code. Then `trim_end_nulls` is fine here. If you deliberately named the test then leave it as it.

uklotzde:
I have also noticed `\0` terminators in `©gen` tags of some .m4a files. Unfortunately, I don't know how they got in there.

src/util/text.rs
uklotzde: ...though renaming the test would be helpful, if I think about it again. Easier to correlate.

Serial-ATA:
They can also show up in Vorbis Comments apparently (https://github.com/Serial-ATA/lofty-rs/issues/273#issuecomment-1773504494). This is something that'll eventually have to be done in other places, it's just most important to get it done for ID3v2 first.

sublipri:
I've changed the test name. Let me know if there are any more issues otherwise I think this is ready to merge.

Serial-ATA:
Thanks for working on this!

PR #284, opened 2023-10-26
ID3v2: Properly handle multi-value UTF-16 encoded frames

### LINKED ISSUE #265: IDv2.4: Parsing multi-valued UTF-16 text fields fails
### Reproducer

Parsing fails with `BadFrameId` errors.

### Summary

https://github.com/gerbera/gerbera/issues/759

> According to http://id3.org/id3v2.4.0-frames ID3v2.4 uses the null character to separate multiple values, which are allowed for all text information frames (tags beginning with T like TEP1 and TCOM).

### Expected behavior

t.b.d.

Might require API changes to handle those multiple values properly.

A lossy implementation that is compatible with the current API could only read the first, non-empty string and silently discard all subsequent strings. While this might be suitable for an application (Example: Mixxx), it is inappropriate for a general purpose library.

### Assets

ID3v2.4 example: 
[txxx_utf16_multi_value_id3v24.zip](https://github.com/Serial-ATA/lofty-rs/files/12909335/txxx_utf16_multi_value_id3v24.zip)

Serial-ATA:
closes #265


uklotzde:
How about releasing a new minor version that includes the latest fixes?

Not sure if #286 is ready for approval yet.

Serial-ATA:
@uklotzde It'll have to be a major release, I'm not too good at keeping breaking changes out of main. :smile:

I haven't looked at #286 yet, but I'd like to get it in before 0.17.0.

PR #290, opened 2023-11-11
Make all file writes generic

Serial-ATA:
This adds two new traits, `Truncate` and `Length`, which are two operations we needed
that required the use of `File`.

With these two traits, combined with `std::io::Read`, `std::io::Write` and `std::io::Seek`, other types can now be treated as files.

It is now possible to write to a `Cursor<Vec<u8>>`.

TODO: changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 	  Lofty writes tags. These are best used as global user-configurable options, as most options will
 	  not apply to all files. The defaults are set to be as safe as possible,
 	  see [here](https://docs.rs/lofty/latest/lofty/struct.WriteOptions.html#impl-Default-for-WriteOptions).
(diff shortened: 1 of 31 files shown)


Serial-ATA:
> Are you interested in any help getting this one over the finish line?

Aside from me having to rebase this since its months old, it actually works just fine. I originally kept this as a draft due to the insane number of repetitive `where` clauses I had to add just to get it working.

I was hoping to find some kind of hack to clean this up, but in the end the only way is just waiting for [Implied Bounds](https://rust-lang.github.io/rfcs/2089-implied-bounds.html) to land (so probably 6 years from now 🫤).

I haven't really looked at this in awhile, but since there's a handful of reactions on it I'll go ahead and merge it in as-is. I'll just need a day or two to bring the branch up to date.

milesegan:
Thanks! I'm happy to help out with any grunt work if that's useful.

Thanks again for all your great work on lofty!

Serial-ATA:
Dealing with all those merge conflicts was painful, I won't let PRs sit like this anymore :smile:.

This'll be out in 0.19, still a few things I need to do before release.

PR #300, opened 2023-12-17
[wav] Fix u32 overflow by first casting stream_len to u64 before multiplying

obsoleszenz:
This caused a panic when trying to probe a >60min .wav set of me.


Serial-ATA:
Good catch, thanks!

PR #303, opened 2023-12-21
Rust v1.70: Replace once_cell with std::sync::OnceLock

uklotzde:


Serial-ATA:
Honestly, I'd rather wait for `LazyLock` to finally get stabilized (probably going to be at least another year from now, given how slowly things move in std). It's a lot simpler to work with, and lofty has few dependencies anyway.

Unless, of course, there is some advantage to using `OnceLock` that I don't know about?

uklotzde:
> Unless, of course, there is some advantage to using `OnceLock` that I don't know about?

Just one less dependency, no other advantages.

I wasn't aware of the plans for the `LazyLock` wrapper. Sounds like it only adds a more convenient API that avoids some boilerplate for simple cases.

uklotzde:
Using a local, static instance inside the functions doesn't add much boilerplate and keeps the namespace clean.

The migration to `LazyLock`would be straightforward when it becomes available. Then you you also need to bump the MSRV again.

Cargo.toml
```diff
@@ -10,6 +10,7 @@ keywords    = ["tags", "audio", "metadata", "id3", "vorbis"]
 categories  = ["multimedia", "multimedia::audio", "parser-implementations"]
 readme      = "README.md"
 include     = ["src", "Cargo.toml", "LICENSE-APACHE", "LICENSE-MIT", "SUPPORTED_FORMATS.md"]
+rust        = "1.70.0"
```
Serial-ATA, line 13:
```suggestion
```
I'd rather not make any guarantees about the MSRV yet. I always just use the latest stable, so this is something extra I'd have to remember to bump on each release.

Serial-ATA (changes requested):
Sure, this seems alright. I just don't want to specify an MSRV right now.

Serial-ATA:
Thanks!

PR #305, opened 2023-12-22
Replace base64 with data-encoding

uklotzde:
More versatile, easy to use, smaller, well-maintained.


uklotzde:
Just a proposal. We use it in all our projects and never looked back.

Serial-ATA:
Nice, `base64` had some weird changes, making it more complicated than necessary. Especially for the simple cases here.

uklotzde:
It is reassuring that you agree and share my point of view.

PR #329, opened 2024-01-05
Rename `PodcastReleaseDate` to `ReleaseDate`

### LINKED ISSUE #328: Naming: `PodcastReleaseDate`
### Summary

The corresponding file tags are not restricted to podcasts. Why is the item named `PodcastReleaseDate` instead of just `ReleaseDate`? The specific use case for tagging podcast episodes should only be mentioned in the docs.

Since this item was only used for MP3 files until v0.17.1 renaming it now before extending the support for more file types might be the last chance without causing too much trouble and confusion.

### API design

Rename `PodcastReleaseDate` to `ReleaseDate` and document the use case for podcast episodes as one example.

uklotzde:
Fixes #328.

### DIFF
--- a/src/tag/item.rs
+++ b/src/tag/item.rs
@@ -106,7 +106,7 @@ gen_map!(
 	"Track"                        => TrackTotal,
 	"Year"                         => Year,
 	"ORIGINALYEAR"                 => OriginalReleaseDate,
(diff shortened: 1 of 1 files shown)


Serial-ATA:
Huh, good catch. Not sure how I managed to group that under podcasts.

PR #330, opened 2024-01-06
MimeType: Implement Display instead of ToString

uklotzde:


Serial-ATA:
Thanks, been meaning to do this.

PR #331, opened 2024-01-06
vorbis: Fix unreachable pattern in ItemKey mapping

uklotzde:


uklotzde:
Causes a clippy warning that slipped through.

Serial-ATA:
> Why do the CI builds succeed with this warning?

Good catch, I haven't seen this warning locally for some reason. I only check CI logs when something fails.

PR #335, opened 2024-01-10
ItemKey: Distinguish between Bpm and IntegerBpm

### LINKED ISSUE #334: Add another `ItemKey` variant for float BPM values
Right now, MP4 has its integer and float BPM atoms map to the integer BPM atom:

https://github.com/Serial-ATA/lofty-rs/blob/f4be1e3775e233847f1620fb4295302817f95341/src/tag/item.rs#L280

These are different values, and should not collide with each other. There should be an `IntegerBpm` and (`FloatBpm` | `PreciseBpm`) `ItemKey` variant.

I do not know yet if other tag formats have a specified float BPM item.

uklotzde:
Fixes #334.

### DIFF
--- a/src/tag/item.rs
+++ b/src/tag/item.rs
@@ -666,7 +666,7 @@ gen_item_keys!(
 		InitialKey,
 		Color,
 		Mood,
-		/// Decimal BPM value with custom precision
+		/// Decimal BPM value with arbitrary precision
 		///
 		/// Only read and written if the tag format supports a field for decimal BPM values
 		/// that are not restricted to integer values.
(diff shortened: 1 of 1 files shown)


Serial-ATA:
I believe "BPM" in Vorbis Comments and APE is also expected to be an integer value. At least that's how Picard uses it.

uklotzde:
> I believe "BPM" in Vorbis Comments and APE is also expected to be an integer value. At least that's how Picard uses it.

I am only aware of ID3v2 and MP4 which impose these restriction. MP4 by atom type and ID3v2 by [standard definition](https://github.com/id3/ID3v2.4/blob/516075e38ff648a6390e48aff490abed987d3199/id3v2.4.0-frames.txt#L376).

uklotzde:
Client still need to handle case that the integer field contains an arbitrary string, e.g. a decimal number.

Serial-ATA:
Yeah, better to play it safe here.

This is marked draft, was there anything else planned for it?

uklotzde:
> Yeah, better to play it safe here.
> 
> This is marked draft, was there anything else planned for it?

I first wanted to publish this for discussion. If you agree we could merge it.

Serial-ATA:
Yep, these changes are good. Thanks!

PR #338, opened 2024-01-11
id3v2: Add support for TXXX:BPM frame with fractional BPM value

uklotzde:


uklotzde:
I can't find the reference where I have seen this.

Issue #345, opened 2024-01-22
Stream picture tags


probablykasper:
### Summary

The ability to read the picture tags as a stream for improve performance

### API design

```rs
let probe = lofty::Probe::open(path);

let tags_reader = probe.into_reader();

// Consumes tags_reader and returns an iterator that skips non-picture tags
let pictures_iterator = tags_reader.pictures();

let first_picture_reader = pictures_iterator.next()?;

let bytes = picture_reader.data; // std::io::Bytes<> iterator
let mime_type = picture_reader.mime_type;
```

Serial-ATA:
Thanks for the suggestion. Could you explain a little more what you're looking for, or your use case for this? I'm not sure I entirely understand what's being suggested. I interpret it as turn Lofty into an event-based parser like [xml-rs](https://crates.io/crates/xml).

probablykasper:
Yeah I guess it is like xml-rs.

### Use case
For my music player, I'm passing song artworks into a webview to show them. I'd like to make that faster by only parsing the first picture (skipping the other items), and to stream the picture to the webview while it's being parsed.

<details>
<summary>Screen recording</summary>

https://github.com/Serial-ATA/lofty-rs/assets/11315492/68d56822-ed3a-4c33-add8-8a5cd02bbb53

</details>

Another thing I didn't think of is the ability to cancel mid-parsing, that would probably be the most significant improvement.

Serial-ATA:
I have thought about an event-based parser before. Lofty currently doesn't offer much caller control over the parsing process, which I don't like.

I'd like Lofty to be a low level crate that people can build abstractions over. `TaggedFile` and `Tag` could still exist as abstractions that simply retain all information from every event.

This would, however, require a decent sized rewrite of the crate. I'm not opposed to doing that, but there'd need to be a more fleshed out design first.

Some initial thoughts:

> let tags_reader = probe.into_reader();

Since `Probe`, `TaggedFile`, and `Tag` need to be backed by concrete formats, a generic event stream could exist where new items first have their key converted to an `ItemKey`. These events would be separate from keys that cannot be represented as an `ItemKey`. That could finally remove `ItemKey::Unknown`. That'd be great.

Sketch of what I mean:

```rust
enum EventKind {
    Item(ItemKey),
    UnknownItem(String),
}

while let Some(e) = events.next() {
    match e {
        Ok(EventKind::Item(_key)) => e.skip(),
[...]

PR #363, opened 2024-04-03
misc: Introduce WriteOptions

### LINKED ISSUE #228: Add write options
### Summary

Just like `ParseOptions`, there may be some things that a caller wants to configure during a write. These options would be passed to [`AudioFile::save_to`](https://docs.rs/lofty/latest/lofty/trait.AudioFile.html#tymethod.save_to) and [`TagExt::save_to`](https://docs.rs/lofty/latest/lofty/trait.TagExt.html#tymethod.save_to).

Some options I can think of being useful are:
* `preferred_padding`: Right now we *make use* of padding, but in its absence we don't end up adding any. Since this is something multiple tags support, it would be nice to add a specified amount of padding to tags when requested.
* `remove_others`: Removing tags from an existing file isn't the easiest process currently. It would be nice if the caller could specify that they want to, for example, remove an APE and ID3v1 tag when they are only attempting to save an ID3v2 tag.
There are currently three ways to remove tags:
	* [`TagExt::remove_from`](https://docs.rs/lofty/latest/lofty/trait.TagExt.html#tymethod.remove_from)/[`TagType::remove_from`](https://docs.rs/lofty/latest/lofty/enum.TagType.html#method.remove_from)
		* These remove tags one at a time, but that isn't helpful if there a singular preferred tag type to be written, as it would require the caller to remove each tag, as well as seek back to the beginning of the file to then do their write.
	* [`TagExt::save_to`](https://docs.rs/lofty/latest/lofty/trait.TagExt.html#tymethod.save_to) with an empty tag
		* Same issue as above
	* [`AudioFile::save_to`](https://docs.rs/lofty/latest/lofty/trait.AudioFile.html#tymethod.save_to)
		* This will handle the seeking, but will not remove any tag field that is `None`. This means that calling `<AudioFile>::remove_<tag>` will not remove <tag> from a file it is saving to. A workaround for this is to do `<AudioFile>::<tag>_mut` followed by [`TagExt::clear`](https://docs.rs/lofty/latest/lofty/trait.TagExt.html#tymethod.clear), as empty tags *will* be treated as a removal request.

### API design

```rust
pub struct WriteOptions {
    preferred_padding: usize,
    remove_others: bool,
    // ...?
}

pub trait TagExt {
    fn save_to(&self, file: &mut File, options: WriteOptions) -> std::result::Result<(), Self::Err>;
}

pub trait AudioFile {
    fn save_to(&self, file: &mut File, options: WriteOptions) -> Result<()>;
}
```

Serial-ATA:
This allows the caller to tweak how Lofty writes their tags in various ways.

As this is just a dumping ground for all sorts of format-specific settings, this is best used as an application global config that gets set once.

In its current state, it will only respect `uppercase_id3v2_chunk` and `preferred_padding` (for some formats).

`respect_read_only` and `remove_others` are defined for later use.

closes #228


src/write_options.rs
```diff
@@ -0,0 +1,131 @@
+/// Options to control how Lofty writes to a file
+#[derive(Copy, Clone, Debug, Ord, PartialOrd, Eq, PartialEq)]
```
uklotzde, line 2: Making this public type _copyable_ could require breaking API changes in the future if it no longer implements `Copy`. If not needed internally then better remove this derive. Could be added later when it is actually needed.

uklotzde: And the `Ord` traits don't make any sense to me.

tests/files/mpeg.rs
```diff
@@ -149,7 +149,7 @@ fn save_to_id3v2() {
 	tag.set_title("title".to_string());
 
 	file.rewind().unwrap();
-	tag.save_to(&mut file).unwrap();
+	tag.save_to(&mut file, WriteOptions::new()).unwrap();
```
uklotzde, line 152: Use `WriteOptions::default()` to make this more explicit? I even prefer the generic `Default::default()` if the type could be inferred.

src/write_options.rs
Serial-ATA: Good catch with `Ord`. I just copied `ParseOptions` and renamed it to `WriteOptions`. Guess I copied it from some other struct. I'll have to remove it from both.

tests/files/mpeg.rs
Serial-ATA: Yeah, `default()` is probably better.

src/write_options.rs
Serial-ATA: > in the future if it no longer implements Copy

I really only intend for `WriteOptions` to be ints and bools just like `ParseOptions`. Since it's used in generated code in `lofty_attr`, it'll just be easier to keep it `Copy` for now.

I'm not too worried about breaking changes at the moment, especially since I have to work on #345 sometime in the near future.

PR #374, opened 2024-04-10
The great Lofty module refactor

Serial-ATA:
Introduce `lofty::prelude` to include all of the common items (`Accessor`, `TagExt`, etc..).

In doing so, I have stopped relying on root re-exports, and come up with a pretty nice module structure.


Serial-ATA:
Also have almost 100 doc tests to fix the imports for.

PR #404, opened 2024-06-05
Add into_inner for BoundTaggedFile

rakbladsvalsen:
Hey, hello there!

I noticed `BoundTaggedFile` (which takes a file) doesn't have a method to retrieve back the original file handle, so I added a fairly simple and short method to retrieve it back. 

I also added a mp3 test case for it.


Serial-ATA:
Thanks!

I changed the test to use the `temp_file!` macro so the file doesn't have to be reverted to its original state.

PR #414, opened 2024-07-03
Add Tag::take_filter()

uklotzde:
The behavior of the `take_string()` and other string accessor methods has changed in lofty v0.20.0. They now return values from all `TagItem`s, regardless of their description. This is unexpected for ID3v2 tags where the description is meaningful. Regular `TagItem`s usually have an empty description.

The new methods are needed to extract only regular `TagItem`s with an empty description from an ID3v2 tag. The example demonstrates how to achieve this for `ItemKey::Comment`.

### DIFF
--- a/lofty/src/tag/item.rs
+++ b/lofty/src/tag/item.rs
@@ -828,16 +828,34 @@ impl TagItem {
 
 	/// Set a language for the [`TagItem`]
 	///
+	/// The default language is empty.
+	///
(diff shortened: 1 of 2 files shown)


Serial-ATA (approved):
Good idea, thanks!

uklotzde:
I have added some comments. Preserving the order of all items while splitting them is an important feature.

Issue #417, opened 2024-07-04
Issues with ID3v2 TRCK count/total separation


basharovV:
### Reproducer

I tried this code:

```rust
fn write_track_number() {
    let mut tag = lofty::tag::Tag::new(TagType::Id3v2);
    tag.insert(TagItem::new(
        ItemKey::from_key(TagType::Id3v2, "TRCK"),
        ItemValue::Text(String::from("1")),
    ));

    println!("{:?}", tag.track()); // None
    println!("{:?}\n", tag.track_total()); // Some(1)

    tag.insert(TagItem::new(
        ItemKey::from_key(TagType::Id3v2, "TRCK"),
        ItemValue::Text(String::from("1/14")),
    ));

    println!("{:?}", tag.track()); // None
    println!("{:?}\n", tag.track_total()); // None

    tag.save_to_path("src/blank.mp3", WriteOptions::new());

    let mp3_file = read_from_path("src/blank.mp3").unwrap();

    for tag_item in mp3_file.tags() {
        for tag in tag_item.items() {
            println!("Found tag: {:?}", tag); // Value is "0/1/14"?
        }
    }
}
```

Output:
```
None
Some(1)

None
None

Found tag: TagItem { lang: [88, 88, 88], description: "", item_key: TrackTotal, item_value: Text("0/1/14") }
```

### Summary

I'm not sure if I'm doing this wrong, but when writing the TRCK tag, I'm seeing some strange behaviour:
- `tag.track()` is `None` and `tag.track_total()` is `Some(1)` when TRCK is a single number. Shouldn't this be the other way round? 
- `tag.track()` and `tag.track_total()` are `None` when providing a delimited TRCK like "1/14"
- A "0/" is prefixed when saving the tag to a file, so the resulting tag looks like "0/1/14" instead of "1/14" (or "0/1" instead of just "1")

Serial-ATA:
Hello!

`ItemKey::from_key` is confusing and really only used internally. I might make it `doc(hidden)` or at least document it better.

So, `ItemKey` doesn't retain any format-specific keys. You should be using `ItemKey::TrackTotal` and `ItemKey::TrackNumber` individually. They will map appropriately when writing the tag.

basharovV:
[...]

PR #431, opened 2024-07-15
Handle edge case for track() parsing

Gusted:
Some audio files also include the total number of tracks in the 'track' tag. This should be ignored in the `track()` parsing. A test is added to verify the new behavior.

### DIFF
--- a/lofty/src/tag/mod.rs
+++ b/lofty/src/tag/mod.rs
@@ -138,8 +138,16 @@ impl Accessor for Tag {
 		Comment     => comment
 	);
 
+	// Some specifications, like IDv2, allow for the track value to be in the
+	// format of x/y, where x is the track position and y is the total number of
+	// tracks. This function only returns the track position and ignores the
+	// total number of tracks.
 	fn track(&self) -> Option<u32> {
-		self.get_u32_from_string(&ItemKey::TrackNumber)
+		let mut i = self.get_string(&ItemKey::TrackNumber)?;
+		if let Some(idx) = i.find('/') {
+			i = &i[..idx];
+		};
+		i.parse::<u32>().ok()
 	}
 
 	fn set_track(&mut self, value: u32) {
--- a/lofty/tests/files/mpeg.rs
+++ b/lofty/tests/files/mpeg.rs
@@ -211,6 +211,40 @@ fn save_number_of_track_and_disk_to_id3v2() {
 }
 
 #[test]
+fn track_edge_case_id3v2() {
+	let mut file = temp_file!("tests/files/assets/minimal/full_test.mp3");
+
+	let tagged_file = Probe::new(&mut file)
+		.options(ParseOptions::new().read_properties(false))
+		.guess_file_type()
+		.unwrap()
+		.read()
+		.unwrap();
+
+	assert_eq!(tagged_file.file_type(), FileType::Mpeg);
+
+	let mut tag = Tag::new(TagType::Id3v2);
+
+	tag.insert_text(ItemKey::TrackNumber, "22/33".to_string());
+
+	file.rewind().unwrap();
+	tag.save_to(&mut file, WriteOptions::default()).unwrap();
+
+	// Now reread the file
+	file.rewind().unwrap();
+	let tagged_file = Probe::new(&mut file)
+		.options(ParseOptions::new().read_properties(false))
+		.guess_file_type()
+		.unwrap()
+		.read()
+		.unwrap();
+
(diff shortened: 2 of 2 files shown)


Serial-ATA:
Hello!

`ItemKey::TrackNumber` is only meant for the current track number, `ItemKey::TrackTotal` is separate.

The `Id3v2Tag` -> `Tag` conversion already splits the TRCK frame's `current/total` format to those two `ItemKey`s. See #417.

Gusted:
Ah sorry, it seems like something on my part went wrong. The audio file in question was using vorbis comments, which doesn't allow this. This patch is indeed redundant, sorry for the noise!

RogueOneEcho:
@Serial-ATA I'm seeing quite a few instances where VORBIS comments have the `current/total` format. This might be worth considering handling

```
comment[12]: TRACKNUMBER=1/1
```

Serial-ATA:
@RogueOneEcho 

Sure, I have no opposition to handling it. It's just that the logic wouldn't go in [`<Tag as Accessor>::track()`](https://docs.rs/lofty/latest/lofty/tag/struct.Tag.html#method.track), but rather the `VorbisComments` parsing, taking into account:

* Are we using [`ParsingMode::Strict`](https://docs.rs/lofty/latest/lofty/config/enum.ParsingMode.html#variant.Strict)? If so, and the format is wrong, error.
* Are [`implicit_conversions`](https://docs.rs/lofty/latest/lofty/config/struct.ParseOptions.html#method.implicit_conversions) enabled? If not, do no special parsing.

So it ends up being a little more work. If you make an issue for it I could see about getting it in for 0.22.0. No guarantee, though.

RogueOneEcho:
I've handled it outside of lofty for my use case so won't submit a PR, but in case somebody searching comes across this here's a solution for handling: 
- total formatted track numbering: `1/10`
- vinyl formatted track numbering: `A1`, `B1`, `B2` etc
```rust
pub(crate) fn fix_track_numbering(tags: &mut Tag) -> bool {
    if tags.track().is_some() {
        return true;
    }
    if replace_total_track_numbering(tags).is_ok() {
        return true;
    }
    if replace_vinyl_track_numbering(tags).is_ok() {
        return true;
    }
    false
}

fn replace_vinyl_track_numbering(tags: &mut Tag) -> Result<(), Error> {
    let track = tags.get_string(&TrackNumber).ok_or_else(|| {
        error(
            "replace vinyl track numbering",
            "No track number string".to_owned(),
        )
    })?;
    let (disc_number, track_number) = get_numeric_from_vinyl_format(track).ok_or_else(|| {
        error(
            "replace vinyl track numbering",
            "Not vinyl format".to_owned(),
        )
    })?;
    trace!(
        "Replacing vinyl track ({track}) with numeric: track {track_number}, disc {disc_number}"
    );
    tags.set_disk(disc_number);
    tags.set_track(track_number);
    Ok(())
}

fn replace_total_track_numbering(tags: &mut Tag) -> Result<(), Error> {
    let track = tags.get_string(&TrackNumber).ok_or_else(|| {
        error(
            "replace total track numbering",
            "No track number string".to_owned(),
        )
    })?;
    let (track_number, track_total) = get_numeric_from_total_format(track).ok_or_else(|| {
        error(
            "replace total track numbering",
            "Not vinyl format".to_owned(),
        )
    })?;
    trace!(
        "Replacing total track numbering ({track}) with numeric: track {track_number}, total {track_total}"
    );
    tags.set_track(track_number);
    tags.set_track_total(track_total);
    Ok(())
}

pub(crate) fn get_numeric_from_vinyl_format(input: &str) -> Option<(u32, u32)> {
    let re = Regex::new(r"^([A-Z])(\d+)$").ok()?;
    let captures = re.captures(input)?;
    let disc_letter = captures.get(1)?.as_str().chars().next()?;
    let track_number: u32 = captures.get(2)?.as_str().parse().ok()?;
    let disc_number = letter_to_number(disc_letter)?;
    Some((disc_number, track_number))
}

pub(crate) fn get_numeric_from_total_format(input: &str) -> Option<(u32, u32)> {
    let re = Regex::new(r"^(\d+)/(\d+)$").ok()?;
    let captures = re.captures(input)?;
    let track_number: u32 = captures.get(1)?.as_str().parse().ok()?;
    let track_total: u32 = captures.get(2)?.as_str().parse().ok()?;
    Some((track_number, track_total))
}
```

Serial-ATA:
Can you make an issue with your solution? I imagine other people could use it. I might get around to supporting this in the future, I just don't have much time to work on Lofty at the moment.

Issue #499, opened 2025-01-08
VorbisComments: Support `current/total` TRACKNUMBER fields


Serial-ATA:
It seems to be a pretty common issue where, rather than using the `TRACKTOTAL` field, the track total will be stored in the `TRACKNUMBER` field as `current/total`, just like APE and ID3v2.

**Workaround:** https://github.com/Serial-ATA/lofty-rs/pull/431#issuecomment-2453458995

The solution is simple, just need to add a special case for `TRACKNUMBER` here: https://github.com/Serial-ATA/lofty-rs/blob/af0ce91105453b573ee480bfc44c26d2c2e4a8c9/lofty/src/ogg/read.rs#L112

Split on `'/'` if present, and insert both a `TRACKNUMBER` and `TRACKTOTAL` item.

RocketRide9:
I will pick this up. I already have local changes, that seem to work. Will double check it, add test and send a PR

PR #500, opened 2025-01-10
VorbisComments: Support current/total TRACKNUMBER fields

RocketRide9:
Closes: #499


lofty/src/ogg/read.rs
```diff
@@ -167,6 +168,32 @@ where
...
+			// Support the case of TRACKNUMBER being equal to current/total
+			k if k.eq_ignore_ascii_case(b"TRACKNUMBER") => {
+				match utf8_decode_str(value) {
+					Ok(value) => {
+						let mut value_split = value.splitn(2, '/');
+						let track_number: Option<u32> =
+							value_split.next().and_then(|b| b.parse().ok());
+						let track_total: Option<u32> =
+							value_split.next().and_then(|b| b.parse().ok());
+						if let Some(n) = track_number {
```
Serial-ATA, line 180: The workaround [here](https://github.com/Serial-ATA/lofty-rs/pull/431#issuecomment-2453458995) also mentions vinyl track numbers. I think in the case that the track number doesn't parse, just insert it into the tag like any other item.

Serial-ATA (changes requested):
Thanks for working on this! Just one suggestion.

lofty/src/ogg/read.rs
RocketRide9: Added vinyl track numbers and another test. Also pushing this tag if track number can't be parsed.

Serial-ATA: Sorry for the confusion! I meant Lofty shouldn't do anything special with vinyl tracks, since I imagine there are weird formats for those out in the wild. In the case that no track number is parsed, just insert it into the tag like normal.

Made that change in [`e961455` (#500)](https://github.com/Serial-ATA/lofty-rs/pull/500/commits/e961455c22736ed033aebf75f57c5671b3c61a8b)

Serial-ATA (approved):

uklotzde:
For these kind of unit tests it is not required to add more test files. The tags to be parsed could be constructed programmatically during the tests and we already use this approach in many cases.

Using actual test files should be reserverd for testing the (binary) deserialization, not the semantic parsing.

RocketRide9:
@uklotzde I couldn't immediately figure out how to do that. I will try to change tests later. Could you point to test where tags are constructed programmatically?

PR #501, opened 2025-01-19
update url in tag_reader example to correct url to ItemKey

Lepidopteran:
I was just looking at your examples for a personal project of mine and saw that the URL to `ItemKey` was invalid


Serial-ATA:
Good catch, thanks!

PR #504, opened 2025-01-24
Fix references to ItemValue, Picture, and PictureType in ApeTag

Lepidopteran:
I was looking around in the `ApeTag` documentation and found the following to be pointing to invalid locations:

- `ItemValue`
- `ItemValue::Binary`
- `Picture::from_ape_bytes`
- `PictureType::Undefined`

I updated them to point to the right locations and tested it on my machine.
Feel free to let me know if any changes need to be made. :)

> NOTE: This is a duplicate of a previous pull request I made, but with a verified the commit to make GitHub happy.

### DIFF
--- a/lofty/src/ape/tag/mod.rs
+++ b/lofty/src/ape/tag/mod.rs
@@ -56,10 +56,10 @@ macro_rules! impl_accessor {
 /// ## Item storage
 ///
 /// `APE` isn't a very strict format. An [`ApeItem`] only restricted by its name, meaning it can use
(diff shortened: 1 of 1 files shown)


Serial-ATA:
Thanks! I thought this would've been caught by `broken_intra_doc_links`, but I guess it doesn't work when the paths are specified like this. :thinking:

I imagine more broken links exist, since a bunch of things were moved around in 0.19.0.

> This is a duplicate of a previous pull request I made, but with a verified the commit to make GitHub happy

Don't worry about it, commits don't need to be signed.

PR #506, opened 2025-01-25
Fix File Type References in Supported File Types Section in Tag Documentation

### LINKED ISSUE #505: Supported File Types Section Produces Invalid List of FileType References for Tag Formats in Documentation
### Reproducer

Look at the "**Supported file types**" section for any tag format. 

Some examples include:

- [ApeTag](https://docs.rs/lofty/latest/lofty/ape/struct.ApeTag.html)
- [Id3v2Tag](https://docs.rs/lofty/latest/lofty/id3/v2/struct.Id3v2Tag.html)
- [Id3v1Tag](https://docs.rs/lofty/latest/lofty/id3/v1/struct.Id3v1Tag.html)
- [VorbisComments](https://docs.rs/lofty/latest/lofty/ogg/struct.VorbisComments.html)

### Summary

I was looking at the documentation for various tags and I've found every tag format, the **Supported file types** section produces invalid references to the various file types

I'm guessing it's because of `Line 19 to 24` in `lofty_tag.rs` in the `lofty_attr` crate:

```rust
fn emit_doc_comment(&self) -> String {
	match self {
		SupportedFormat::Full(path) => format!(
			"* [`FileType::{ft}`](crate::FileType::{ft})\n",
			ft = path.get_ident().unwrap()
		),
		SupportedFormat::ReadOnly(path) => format!(
			"* [`FileType::{ft}`](crate::FileType::{ft}) **(READ ONLY)**\n",
			ft = path.get_ident().unwrap()
		),
	}
}
```

It's generating documentation that points to `crate::FileType::<File_Format>` which is invalid because, file types are now stored `crate::file::FileType::<File_Format>`.

I did try to change it myself, but when I tested it locally, it did not produce a different documentation result for me.

I'm not sure why I couldn't get it to work, but I wanted to point this out!

### Expected behavior

In the **Supported file types** section, produce a list that points to `crate::file::FileType::<File_Format>` instead of `crate::FileType::<File_Format>`

#### Side Note

This is a amazing project!

I was previously using [`id3`](https://crates.io/crates/id3), [`metaflac`](https://crates.io/crates/metaflac) and [`mp4ameta`](https://crates.io/crates/mp4ameta) for a personal project of mine and I wanted to support reading and writing to more tag formats other than `id3`, `Itunes`, and a *partial* of `vorbiscomments`, and I was daunting the task of writing my own readers and writers.

So when finding your project, I was relieved, and after trying your project out, I'd say you have the best (and to my knowledge only) library that provides a nice API for every popular metadata standard out there! 

So to end all, thank you!

### Assets

_No response_

Lepidopteran:
This fixes #505

I went an implemented the fix suggested in the said issue.

I also changed referencing `lofty_attr` to use `path` instead of published version in `lofty`, 
If this is not wanted, I can easily revert it.

Let me know if this is good to go!

### DIFF
--- a/lofty/Cargo.toml
+++ b/lofty/Cargo.toml
@@ -18,7 +18,7 @@ byteorder     = { workspace = true }
 # ID3 compressed frames
 flate2        = { version = "1.0.30", optional = true }
 # Proc macros
-lofty_attr = "0.11.0"
+lofty_attr = { path = "../lofty_attr" }
 # Debug logging
 log           = "0.4.22"
 # OGG Vorbis/Opus
--- a/lofty_attr/src/lofty_tag.rs
+++ b/lofty_attr/src/lofty_tag.rs
@@ -17,11 +17,11 @@ impl SupportedFormat {
 	fn emit_doc_comment(&self) -> String {
 		match self {
 			SupportedFormat::Full(path) => format!(
-				"* [`FileType::{ft}`](crate::FileType::{ft})\n",
+				"* [`FileType::{ft}`](crate::file::FileType::{ft})\n",
 				ft = path.get_ident().unwrap()
 			),
 			SupportedFormat::ReadOnly(path) => format!(
-				"* [`FileType::{ft}`](crate::FileType::{ft}) **(READ ONLY)**\n",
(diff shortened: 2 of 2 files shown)


Serial-ATA:
Looks good, thanks!

PR #510, opened 2025-02-21
Add ext functions to file types

giraffekey:
Adds two simple quality of life functions to `FileType` and `MimeType`.

```rust
FileType::Mpeg.ext(); // Some("mp3")
MimeType::Jpeg.ext(); // Some("jpg")
```

My use case is that I want to extract the pictures and save them separately to disk, so this saves me a manual pattern match on the extension of the file.

### DIFF
--- a/lofty/src/file/file_type.rs
+++ b/lofty/src/file/file_type.rs
@@ -99,33 +99,6 @@ impl FileType {
 		}
 	}
 
-	/// Returns the extension for the `FileType` if it is known
-	///
-	/// # Examples
-	///
-	/// ```rust
-	/// use lofty::file::FileType;
-	///
-	/// assert_eq!(FileType::Mpeg.ext(), Some("mp3"));
-	/// ```
-	pub fn ext(&self) -> Option<&str> {
-		match self {
-			FileType::Aac => Some("aac"),
-			FileType::Aiff => Some("aiff"),
-			FileType::Ape => Some("ape"),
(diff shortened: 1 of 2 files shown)


lofty/src/file/file_type.rs
```diff
@@ -99,6 +99,33 @@ impl FileType {
...
+			FileType::Mpc => Some("mpc"),
+			FileType::Opus => Some("opus"),
+			FileType::Vorbis => Some("ogg"),
+			FileType::Speex => Some("spx"),
+			FileType::Wav => Some("wav"),
+			FileType::WavPack => Some("wv"),
+			FileType::Custom(_) => None,
+		}
+	}
+
```
Serial-ATA, line 128:
```suggestion
```
I don't think `FileType` should have this method. While these are the most common extensions for these formats, it's possibly information loss.

For example:
* The `aiff` extension can also be `aifc` for compressed audio
* The `mp4` extension doesn't really say much about the file, `m4a`, `m4b`, and `m4v` are all common extensions (with many others being out there).

Serial-ATA (changes requested):
Thanks for the PR!

I think having this on `MimeType` is fine, but it probably shouldn't be on `FileType`.

giraffekey:
Makes sense.

Serial-ATA (approved):

Serial-ATA:
Thanks!

PR #512, opened 2025-03-29
Fix Rust Edition 2024 compatibility issues

uklotzde:


uklotzde:
@Serial-ATA What about the failing Style check that seems to be unrelated?

Serial-ATA:
> @Serial-ATA What about the failing Style check that seems to be unrelated?

I was able to format it on my end. CI uses the latest stable, maybe yours is out of date if you don't see the same formatting?

Cargo.toml
```diff
@@ -1,24 +1,38 @@
...
 explicit_outlives_requirements = "deny"
 unknown_lints                  = "allow"
 
+# Opt-in to new drop order of Edition 2024.
+# TODO: Remove if no longer needed.
+tail_expr_drop_order           = "allow"
+if_let_rescope                 = "allow"
+
+# TODO: Remove after switching to Edition 2024.
+edition_2024_expr_fragment_specifier = "allow"
```
Serial-ATA, line 34: Are these still needed?

uklotzde: Yes. Only after switching to `edition = "2024` both `if_let_rescope` and `edition_2024_expr_fragment_specifier` will probably become obsolete.

Serial-ATA: You can bump to 2024 edition in this PR if you want. I don't think it's a breaking change.

uklotzde:
> > @Serial-ATA What about the failing Style check that seems to be unrelated?
> 
> I was able to format it on my end. CI uses the latest stable, maybe yours is out of date if you don't see the same formatting?

I forgot to run `cargo fmt` manually.

I am used to `pre-commit` for running various checks. Code formatting is only one of them. Produces the same results locally and on CI. At least if the toolchains are in-sync.

Cargo.toml
uklotzde: Switching to Edition 2024 requires to reformat many source files.

I will prepare an intermediate step with some workspace cleanup as part of this PR. Switching editions should be done separately.

Serial-ATA:
Thanks!

PR #542, opened 2025-08-18
VorbisComments: Parse `TRACKNUMBER` with respect to `implicit_conversion`

### LINKED ISSUE #540: VorbisComments: TRACKNUMBER=`current/total` handling should be under implicit conversions
The special case for `TRACKNUMBER` in `VorbisComments` is an implicit conversion, and it can alter metadata even if it isn't in the `current/total` form:

https://github.com/Serial-ATA/lofty-rs/blob/09a9352e418b17868e73e707a015a7609d6f5aac/lofty/src/ogg/read.rs#L171-L203

Should be skipped if [`ParseOptions::implicit_conversions`](https://docs.rs/lofty/latest/lofty/config/struct.ParseOptions.html#method.implicit_conversions) is disabled.

### Discussed in https://github.com/Serial-ATA/lofty-rs/discussions/539

<div type='discussions-op-text'>

<sup>Originally posted by **Methapon2001** August 14, 2025</sup>
When I try to get track number and track total, disc and disc total from my FLAC (or other file type) file which should have leading 0, instead I got number with non leading using `get` or `get_string`.

I tried with wav file it works as expected.

I tried change probed options(disable implicit conversion / relaxed) and it is not working.

I wonder if this is bug or working as expected.

I tried symphonia it does contains zero leading and also in MusicBrainz Picard.</div>

Methapon2001:
# Summary

Fixes #540.
This PR make VorbisComments parse `TRACKNUMBER` with respect to `ParseOption::implicit_conversions`

# Note

Let me know if there is something more that should be done.

I'm not sure if I need to update the `Accessor` for `VorbisComments`. The goal is to ensure that `track` and `track_total` can still be called and return `u32` values, even when the format is `<current>/<total>`. 

```rust
	fn track(&self) -> Option<u32> {
		if let Some(item) = self
			.get("TRACKNUMBER")
			.map_or_else(|| self.get("TRACKNUM"), Some)
		{
			return item.split('/').next().and_then(|v| v.parse::<u32>().ok());
		}

		None
	}
```

What do you think?


Serial-ATA:
Thanks!

> The goal is to ensure that track and track_total can still be called and return u32 values, even when the format is `<current>/<total>`.

Hm, I'm not sure how I feel about `Accessor` methods handling multiple data formats like that. Generally, I prefer modifying the data at *parse time*, and having `Accessor` assume that's the case. That's what `implicit_conversions` warns about in the docs: https://docs.rs/lofty/latest/lofty/config/struct.ParseOptions.html#method.implicit_conversions

Methapon2001:
> Thanks!
> 
> > The goal is to ensure that track and track_total can still be called and return u32 values, even when the format is `<current>/<total>`.
> 
> Hm, I'm not sure how I feel about `Accessor` methods handling multiple data formats like that. Generally, I prefer modifying the data at *parse time*, and having `Accessor` assume that's the case. That's what `implicit_conversions` warns about in the docs: https://docs.rs/lofty/latest/lofty/config/struct.ParseOptions.html#method.implicit_conversions

All right, then the current behaviors should be as expected. I also added 2 test cases.

Serial-ATA:
I went ahead and removed the test files, for a simple case like this it's best to just create the tag in the test

PR #547, opened 2025-09-19
ogg_pager: Fix panic on nil packet

WorkingRobot:
I had a pretty weird crash that gets caused by the old `assert!(packet.is_empty());` assertion. At the assertion point, I had the following state:
`packet.len() = 3908`
`ctx = { ... abgp: 0, stream_serial: 0, header_flags: 2, flags: PaginateContextFlags { first_page: false, fresh_packet: true, packet_finished_on_page: false, need_nil_page: true }, pos: 0, idx: 24, remaining_page_size: 8160, current_packet_len: 3908, last_segment_size: Some(83)`

This PR seems to fix it, but I'm unfamiliar with all the inner workings of ogg, so please let me know if there's something I should fix.

### DIFF
--- a/ogg_pager/CHANGELOG.md
+++ b/ogg_pager/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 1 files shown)


Serial-ATA:
Interesting. Thanks for the fix!

If you still have it, could you email the file to `serial AT [domain on my profile]`? There's already a test in `ogg_pager` for nil packets, but I guess there's some other case I didn't think of.

Issue #549, opened 2025-09-28
Corrupting flac files


phayes:
### Summary

I've been running into issues with lofty corrupting flac files in the following situations:

1. It's a "bare" flac file (eg STREAMINFO is the final block, no tag metadata, no SEEKTABLE), 

2. There's no padding or metadata, just a STREAMINFO and SEEKTABLE. 

I'm not *entirely* sure of my understanding of the bug.  I've laid out below what I think is the problem, but I dont have enough confidence in my understanding to offer a PR:

1. Right here (https://github.com/Serial-ATA/lofty-rs/blob/2727c5ce65501dc57075e3995958ecdea1ea4569/lofty/src/flac/write.rs#L83). `last_block_info` is later used as (header_first_byte, header_offset, payload_end).
You’re overwriting the header offset with (end - start) (the length), and the “header first byte” with block.byte (which is just the value of the first header byte, not an offset). This causes `file_bytes[last_block_info.1] = first_byte;` (https://github.com/Serial-ATA/lofty-rs/blob/2727c5ce65501dc57075e3995958ecdea1ea4569/lofty/src/flac/write.rs#L127) to write the “is_last” flag into a length-sized index. The causes file corruption when we need to adding a padding block. 

2. If there are no blocks to remove, we corrupt the file right here (https://github.com/Serial-ATA/lofty-rs/blob/2727c5ce65501dc57075e3995958ecdea1ea4569/lofty/src/flac/write.rs#L152) `file_bytes.splice(0..0, comment_blocks);` overwrites the first 4 `fLaC` bytes in the file.

phayes:
I've opened a PR for this here: #550 - I would appreciate a careful review since I'm not 100% confident I'm following all the various accouting being done to keep track of things.

KRTirtho:
@Serial-ATA facing the same issue in Spotube (v5.1.0 latest) which uses `lofty-rs` for tagging. You can actually check the downloaded FLAC files with `ffprobe` and it shows:

> Btw, use the `DAB Music` plugin from the plugins page which provides FLAC streaming support`

```
[...]

PR #550, opened 2025-09-28
Fixing corrupting of FLAC files where there is no padding or no metada blocks

phayes:
This is the companion PR to https://github.com/Serial-ATA/lofty-rs/issues/549

I'm not feeling entirely confident that I'm following all the various accouting being done here to keep track of everything. 

I *think* there's still an outsstanding bug where we're not setting the last flag correctly if we have no padding and are removing all picture blocks. 

However, if you review it and you think I have it correct, this should fix the FLAC corruption problems I've been seeing.

### DIFF
--- a/lofty/src/flac/write.rs
+++ b/lofty/src/flac/write.rs
@@ -80,7 +80,8 @@ where
 		last_block = block.last;
 
 		if last_block {
-			last_block_info = (block.byte, (end - start) as usize, end as usize)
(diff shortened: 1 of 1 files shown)


Serial-ATA:
Hm, yeah I don't think any of the FLAC tests have a file with just a single metadata block. I'll test it out in a couple days when I have time. Thanks!

Serial-ATA:
Finally got around to this in #583, thanks for working on this!

PR #552, opened 2025-10-04
fix: replace paste with maintained fork pastey

open-schnick:
As seen in [RUSTSEC-2024-0436](https://rustsec.org/advisories/RUSTSEC-2024-0436), [paste](https://github.com/dtolnay/paste) is no longer maintained and users should switch to [pastey](https://github.com/as1100k/pastey) instead.  
The authers of pastey suggested this change as it does not involve changing the source code. I can also change the source code instead if you want.

### DIFF
--- a/lofty/Cargo.toml
+++ b/lofty/Cargo.toml
@@ -25,7 +25,7 @@ log           = "0.4.22"
 # OGG Vorbis/Opus
 ogg_pager     = "0.7.0"
 # Key maps
-paste         = "1.0.15"
+paste         = { package = "pastey" , version = "0.1.1" }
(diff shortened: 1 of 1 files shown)


uklotzde:
Even if `paste` is no longer maintained. It does not cause any issues and using it is safe.

There is no official recommendation to switch to `pastey`. `pastey`  just _claims_ to be a drop-in replacement.

Why replace a proven crate from a well-known and respected author in the Rust community with a fresh fork? I suggest to defer the switch until either `paste` causes build issues or until other major crates switch from `paste` to `pastey`.

Issue #66, opened 2022-09-04
Support multiple field values (e.g. multiple artists)


UgnilJoZ:
### Summary

[Ogg Vorbis I format specification](https://www.xiph.org/vorbis/doc/v-comment.htmll#fieldnames) mentions multiple possible values per field name:

> Field names are not required to be unique (occur once) within a comment header. As an example, assume a track was recorded by three well know artists; the following is permissible, and encouraged: 
> ```
> ARTIST=Dizzy Gillespie 
> ARTIST=Sonny Rollins
> ARTIST=Sonny Stitt 
> ```

The [Accessor trait](https://docs.rs/lofty/latest/lofty/trait.Accessor.html#method.artist) handles the artist as `String` and `&str`, which is a single value, or sometimes possibly a string that represents a list using a separator.

It seems to be currently impossible to get a list of the artists embedded in Vorbis Comments.

### API design

```rust
impl Accessor {
    pub fn artists(&self) -> impl Iterator<Item=&str>;
    /// Replace the list of artists
    pub fn set_artists(&mut self, values: impl Iterator<Item=&str>);
}
```

Serial-ATA:
This is possible when using `Tag`, with either [get_strings](https://docs.rs/lofty/latest/lofty/struct.Tag.html#method.get_strings) or [take_strings](https://docs.rs/lofty/latest/lofty/struct.Tag.html#method.take_strings). I did miss this on `VorbisComments` though 😄.

open-schnick:
Hi,
i just got it to work using `get_items` for reading and `push` for writing the tags but needless to say it was tiresome to find out how to do that because its not very intuitive.
Are there any plans to adapt the api to reflect the fact that artist can be more than a single one?
If that is not the case i would like to add this "special" case to the docs of the set and get artist functions

Serial-ATA:
@open-schnick

> Are there any plans to adapt the api to reflect the fact that artist can be more than a single one?

I don't really think it's an API issue. 

Note that technically *any* item can appear multiple times. And I don't think changing the API to just return all items would be a great idea, since most users only care about the first item anyway.

[...]

PR #554, opened 2025-10-11
docs(lofty): add information on how to set multiple elements per tag

open-schnick:
See  #66


Serial-ATA:
Thanks! Reworded it a bit in #556

PR #558, opened 2025-10-31
Adding list of extensions

### LINKED ISSUE #509: Add a list of extensions
### Summary

A lot of projects that use Lofty are scanning directories and filtering files based on a set of extensions. What typically ends up happening is only a few popular extensions (mp3, flac, wav) will make it in the list, and not all of the formats that Lofty supports. Not to mention it's just a bunch of duplicated work.

`FileType::from_ext` has a pretty good list of extensions:
https://github.com/Serial-ATA/lofty-rs/blob/aa1ec31ea6f2d6f08cc034cea6aad50923fc5f07/lofty/src/file/file_type.rs#L130-L145

It too isn't perfect, of course. But if projects start using a list of extensions we provide, then they get any improvements to it for free.

### API design

```rust
/// ...
/// NOTE: This is not an exhaustive list, blah blah blah
pub const EXTENSIONS: &[&str] = &["mp3", "flac", ...];
```

kdeng00:
Creating a list of extensions commonly used throughout the project.

Closes #509


Serial-ATA:
Thanks! I updated the docs and changelog.

Issue #559, opened 2025-11-03
Replace unmaintained crate with drop-in replacement, regarding "paste"


Collbrothers:
To prevent any future vulnerabilities, loss of new features & most importantly to allow lofty-rs dependent crates to enforce a strict deny unmaintained dependencies. 

Will open a related PR asap 👍🏻

PR #560, opened 2025-11-03
Replace unmaintained "paste" with drop-in replacement "pastey"

Collbrothers:
Changes are mostly changing "paste::" to "pastey::" & updating Cargo.toml.  

Related to #559.


uklotzde:
#552

uklotzde:
- `paste` is only used at compile-time.
- `paste` is still more widely used.
- No need to replace it with a crate of unknown reputation and future.

Serial-ATA:
Yeah, the maintenance status of `paste` isn't that important here. The crate's been feature complete for years at this point. I'd really only want to switch to `pastey` if I actually need the new features from it.

PR #573, opened 2025-11-12
Add country release

UnknownSuperficialNight:
Add `ReleaseCountry` to `ItemKey`

I propose adding `ReleaseCountry` to `ItemKey`. Please review to ensure I've covered everything.

This change adds `ItemKey` to a few `key/ItemKey maps`.

I have a question regarding `ID3V2_MAP`. Should it be:
```
"TXXX:MusicBrainz Album Release Country" => ReleaseCountry 
OR
"TXXX" => ReleaseCountry
```

I assume the `TXXX:MusicBrainz Album Release Country` one is correct?

I'm also unsure whether to map `MusicBrainz Album Release Country` directly to `ReleaseCountry` or create a separate `MusicBrainzReleaseCountry`. My assumption is to map it to a single `ReleaseCountry`.

Let me know if I missed anything, other, than those two.

Note-to-self:
Todo: ~`ILST_MAP`~ and ~`ID3V2_MAP`~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     - In formats like Vorbis Comments, `ItemKey::Lyrics` may actually contain synchronized lyrics in LRC format. To help with the ambiguity, some
       apps may write a separate field containing normal, unsynchronized lyrics.
     - In other formats where the difference doesn't matter (like ID3v2), this will act exactly the same as `ItemKey::Lyrics`.
+  - `ItemKey::ReleaseCountry` ([PR](https://github.com/Serial-ATA/lofty-rs/pull/573))
+    - Currently, this maps to the [fields](https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html#id30) used by MusicBrainz Picard, which expect an
+      ISO 3166-1 code.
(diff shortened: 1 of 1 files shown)


Serial-ATA:
> I have a question regarding ID3V2_MAP. Should it be:

For keys that map to `TXXX` frames, you just put the description in there, no `TXXX`. So in this case, `"MusicBrainz Album Release Country" => ReleaseCountry`.

Then you just stick the `ItemKey::ReleaseCountry` in this list:

https://github.com/Serial-ATA/lofty-rs/blob/0805a4cf2d4cfc4c4adb0a029f54def0cef7968b/lofty/src/id3/v2/tag.rs#L1415-L1423

Note that only ID3v2 has that special requirement, just because those handful of items map to a frame *other* than what their name specifies.

> I'm also unsure whether to map MusicBrainz Album Release Country directly to ReleaseCountry or create a separate MusicBrainzReleaseCountry. My assumption is to map it to a single ReleaseCountry.

Yeah, it's unfortunate that they stuck "MusicBrainz" in there, but at the same time I *believe* it's the only standardized field for release country. I'd be fine just mapping it to `ReleaseCountry`, cause at the end of the day it's just a standard ISO 3166-1 code, typically.

And if it ever needs to change down the line, it can. No big deal.

UnknownSuperficialNight:
@Serial-ATA Alright have a look if that's correct

Serial-ATA:
Looks good, thanks for keeping the ilst map aligned as well. :)

Rebased and updated the changelog.

PR #576, opened 2025-11-13
ID3v2: Unify generic conversions

### LINKED ISSUE #349: Can not write multiple artists with mp3 and generic `Tag`
### Reproducer

I tried this code:

```rust
let mut tag = Id3v2Tag::default();
tag.set_artist(artists.join("\0")); // ["Artist1", "Artist2"]
let mut tag: Tag = tag.into();
tag.save_to_path(&path).unwrap();

// re-read
let tagged_file = read_from_path(&path).unwrap();
let artists = tagged_file.primary_tag().unwrap().
            .get_strings(&ItemKey::TrackArtist)
            .map(std::string::ToString::to_string)
            .collect_vec(); // ["Artists2"]
```

while this code works:

```rust
let mut tag = Id3v2Tag::default();
tag.set_artist(artists.join("\0")); // ["Artist1", "Artist2"]
tag.save_to_path(&path).unwrap();

// re-read
let tagged_file = read_from_path(&path).unwrap();
let artists = tagged_file.primary_tag().unwrap().
            .get_strings(&ItemKey::TrackArtist)
            .map(std::string::ToString::to_string)
            .collect_vec(); // ["Artist1", "Artist2"]
```

And it seems that this only happens with `mp3`. I can write multiple artists with vorbis comment on a `flac` file.

### Summary

When converting to generic `Tag` from `Id3v2Tag`, multiple artitsts are not preserved.

### Expected behavior

When converting to generic `Tag` from `Id3v2Tag`, multiple artitsts are preserved.

### Assets

_No response_

Serial-ATA:
`Tag::save_to()` and `Into::<Id3v2Tag>::into(tag)` would take different paths, causing frames to not get merged (and possibly other weird behavior). Now, both take the same path.

All frames have been updated to use `Cow`, to avoid allocations in the `Tag::save_to()` path (unless absolutely necessary, like for track numbers).

This should be easier to maintain now that everything's unified. And possibly more efficient, since it's no longer cloning everything.

closes #349


uklotzde:
After writing `ItemKey::IntegerBpm` as ID3v2 into an empty file no tags are written and the file remains empty.

```
let mut file = copy_named_temp_file("tests/assets/empty.mp3");

  {
      let tagged_file = Probe::new(&mut file.as_file())
          .guess_file_type()
          .unwrap()
          .read()
          .unwrap();
      assert_eq!(FileType::Mpeg, tagged_file.file_type());
      assert!(tagged_file.tags().is_empty());
  }

  let mut tag = Tag::new(TagType::Id3v2);
  tag.insert_text(
      ItemKey::IntegerBpm,
     122.to_string(),
  );
  tag.save_to_path(file.path(), WriteOptions::default())
      .unwrap();
```

uklotzde:
https://github.com/aoide-org/aoide-rs/actions/runs/19388551698/job/55478819021

Serial-ATA:
@uklotzde Good catch, fixed in #579.

PR #590, opened 2026-01-14
ID3v2: Fix parsing WFED as UrlLinkFrame

### LINKED ISSUE #589: Problem with parsing WFED tag in id3v2
### Reproducer

I'm reading a file through Probe, but I get the error TextDecode("Found invalid encoding")

[01. Red Beast.mp3](https://github.com/user-attachments/files/24614167/01.Red.Beast.mp3)

### Summary

After investigating the issue, I discovered that mutagen in Python reads correctly. After narrowing down the search, I found the problem with WFED. I'm new to this and can't pinpoint the cause, but lofty doesn't find any encodings for this tag because encoding_byte > 100.

I discovered that if WFED is read as a URL Frame, everything works correctly.

```rust
// WFED (Podcast URL), GRP1 (Grouping), MVNM (Movement Name), MVIN (Movement Number)
"GRP1" | "MVNM" | "MVIN" => TextInformationFrame::parse(reader, id, flags, version)?.map(Frame::Text),
"WFED" => UrlLinkFrame::parse(reader, id, flags)?.map(Frame::Url),
```

Image

I think create PR, but I'm new to Rust, and especially to mp3 tags. But maybe you could look into this issue and the suggested fix.

From mutagen debug
Image

### Expected behavior

_No response_

### Assets

_No response_

NikkoFox:
close #589

### DIFF
--- a/lofty/src/id3/v2/frame/content.rs
+++ b/lofty/src/id3/v2/frame/content.rs
@@ -42,7 +42,7 @@ pub(super) fn parse_content<R: Read>(
 		i if i.starts_with('T') => TextInformationFrame::parse(reader, id, flags, version)?.map(Frame::Text),
 		// Apple proprietary frames
 		// WFED (Podcast URL), GRP1 (Grouping), MVNM (Movement Name), MVIN (Movement Number)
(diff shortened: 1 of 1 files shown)


Serial-ATA:
Thanks!

PR #591, opened 2026-01-15
ID3v1: Fix resize_string to safely split utf-8 chars

NikkoFox:
Incorrect UTF-8 string truncation causes panic

The code truncates strings by byte index, assuming that cutting at a fixed byte length is always safe.
However slicing at an arbitrary byte offset may split a multibyte character, which results in a runtime panic.

**Panic example**

```rust
fn main() {
    let s = "На московских... окраинах..."; // UTF-8, Cyrillic characters
    println!("{:?}", &s.split_at(30).0.as_bytes());
}
```

The new version of rust has a suitable method and I emulated its work for splitting UTF-8 characters

https://doc.rust-lang.org/src/core/str/mod.rs.html#410

```rust
fn main() {
    let s = "На московских... окраинах..."; // UTF-8, Cyrillic characters
    let mut end = 30;
    while !s.is_char_boundary(end) {
        end -= 1;
    }
    println!("{:?}", &s[..end].as_bytes());
}
```

### DIFF
--- a/lofty/src/id3/v1/write.rs
+++ b/lofty/src/id3/v1/write.rs
@@ -50,13 +50,18 @@ where
 }
 
 pub(super) fn encode(tag: &Id3v1TagRef<'_>) -> std::io::Result<Vec<u8>> {
+	#[allow(clippy::sliced_string_as_bytes)]
 	fn resize_string(value: Option<&str>, size: usize) -> std::io::Result<Vec<u8>> {
 		let mut cursor = Cursor::new(vec![0; size]);
 		cursor.rewind()?;
 
 		if let Some(val) = value {
 			if val.len() > size {
-				cursor.write_all(val.split_at(size).0.as_bytes())?;
+				let mut end = size;
+				while !val.is_char_boundary(end) {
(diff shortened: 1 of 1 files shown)


Serial-ATA:
ID3v1 text fields are Latin-1, not UTF-8. `encode` should either fail if the string isn't representable in Latin-1 or do a lossy re-encode. Could even make it possible to choose between the two in `WriteOptions`.

NikkoFox:
Am I correct in understanding that panic, in this case, is the expected behavior? 
Since no conversion to Latin1 occurs here, yet Latin1 is the anticipated single-byte encoding.

Serial-ATA:
Panicking isn't the correct behavior here, the Latin-1 encoding was just overlooked. It definitely should be fixed. ID3v1 isn't used that much, so it just wasn't caught before. All of the tests also only use ASCII strings.

Serial-ATA:
Fixed in #594, which adds `WriteOptions::lossy_text_encoding()`. By default, invalid characters will be replace with `?`. Disabling the option will return an error when encoding.

PR #594, opened 2026-01-18
WriteOptions: Add `lossy_text_encoding` option

Serial-ATA:
Text encoding is now fallible, with the option to replace invalid characters with `'?'` or return an error. For now, the only fallible encoding is `TextEncoding::Latin1`.


NikkoFox:
In my case, there’s no more panic, thank you

PR #599, opened 2026-02-03
MP4: Fix freeform atom size validation

M0Rf30:
Change freeform chunk size check from >= to > to allow atoms that exactly fill the remaining space

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     header, then the parser would error.
 - **FLAC**: Fix corruption of files with no metadata blocks ([issue](https://github.com/Serial-ATA/lofty-rs/issues/549)) ([PR](https://github.com/Serial-ATA/lofty-rs/pull/583))
 - **OGG**: Improved performance of page searching ([issue](https://github.com/Serial-ATA/lofty-rs/issues/588)) ([PR](https://github.com/Serial-ATA/lofty-rs/pull/600))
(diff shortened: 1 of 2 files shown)


M0Rf30:
## Issue
M4A files with embedded cover art fail to extract artwork. The parser stops prematurely when encountering iTunes freeform atoms (----), never reaching the covr (cover art) atom.

## Root Cause
In lofty/src/mp4/atom_info.rs:276, the freeform chunk size validation incorrectly rejects atoms whose size exactly matches the remaining buffer space:
```rust
// Before (incorrect)
if len >= *reader_size {
    err!(SizeMismatch);
}
// After (correct)  
if len > *reader_size {
    err!(SizeMismatch);
}
```
## Fix

Changed >= to > to allow atoms that perfectly fill remaining space, which is valid per MP4 spec.

## Testing
- Tested with M4A files containing iTunes metadata and embedded cover art
- Parser now successfully processes all atoms including covr
- Cover art extraction works correctly

M0Rf30:
Any news about this?

Serial-ATA:
Hey @M0Rf30, this makes sense. Just curious, do you have a file that triggers this? I don't think I've ever encountered a freeform atom without data

M0Rf30:
> Hey @M0Rf30, this makes sense. Just curious, do you have a file that triggers this? I don't think I've ever encountered a freeform atom without data

of course
[01.zip](https://github.com/user-attachments/files/25160683/01.zip)

set the extension of downloaded file back to m4a

Serial-ATA:
Ah interesting, it just has a bunch of holes in the data. I was expecting some flag atoms. Still valid in any case.

PR #614, opened 2026-02-15
ID3v2: fallback when BOM is not present in UTF-16 data

### LINKED ISSUE #613: Parsing mp3 file fails when UTF-16 data has missing BOM
### Reproducer

_No response_

### Summary

#535 caused a regression in some mp3 files that don't include a valid BOM in a language frame or synchronized text frame. [Here](https://github.com/Serial-ATA/lofty-rs/blob/95e168ebca466e085fe61393fdbb7a74069c2608/lofty/src/id3/v2/items/language_frame.rs#L58-L60) it returns an error if the BOM is missing which causes the entire file to fail to parse. Before #535, files that exhibited this behavior still parsed successfully.

### Expected behavior

Files with missing BOMs should still parse successfully.

### Assets

_No response_

aschey:
Resolves #613

This restores the behavior prior to #535 where language frames are still parsed successfully if the BOM is not present in UTF-16 data. I don't have any files that cause an issue here with text frames, but I updated it there too for consistently. I can revert that if desired.

### DIFF
--- a/lofty/src/id3/v2/items/language_frame.rs
+++ b/lofty/src/id3/v2/items/language_frame.rs
@@ -3,7 +3,6 @@ use crate::error::{Id3v2Error, Id3v2ErrorKind, Result};
 use crate::id3::v2::frame::content::verify_encoding;
 use crate::id3::v2::header::Id3v2Version;
 use crate::id3::v2::{FrameFlags, FrameHeader, FrameId};
-use crate::macros::err;
 use crate::tag::items::Lang;
 use crate::util::text::{
 	DecodeTextResult, TextDecodeOptions, TextEncoding, decode_text,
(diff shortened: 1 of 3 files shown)


lofty/src/id3/v2/items/language_frame.rs
```diff
@@ -49,20 +48,20 @@ impl LanguageFrame {
...
-		}
+		let endianness: Option<fn([u8; 2]) -> u16> = if encoding == TextEncoding::UTF16 {
+			match bom {
+				[0xFF, 0xFE] => Some(u16::from_le_bytes),
+				[0xFE, 0xFF] => Some(u16::from_be_bytes),
+				_ => None,
+			}
+		} else {
+			None
+		};
```
Serial-ATA: Do you have an example with a broken language frame? Not sure I understand what the solution here is. `decode_text` with an encoding of `UTF16` should fail anyway if there isn't a valid BOM.

aschey: Sure, here's one that reproduces the problem. The frame does decode correctly with this change in place. I don't know much about id3 tags so maybe this is just a band-aid fix, but I was trying to preserve the behavior from 0.22 where this worked previously.

[04 Blood on the Radio.zip](https://github.com/user-attachments/files/25327210/04.Blood.on.the.Radio.zip) - the lyrics tag is the one that fails here without this change

Serial-ATA: Ah, yeah. #535 broke the case of _empty_ descriptions which shouldn't have a BOM. This change makes sense then, I'll just update the comment

Serial-ATA:
Thanks!

PR #625, opened 2026-03-05
fix: replace abandoned paste crate with pastey (RUSTSEC-2024-0436)

boul2gom:


naglis:
See also #560

Serial-ATA:
https://github.com/Serial-ATA/lofty-rs/pull/552#issuecomment-3368495075

There's no real need to switch to a maintained fork of `paste`, the crate is simple and already suits Lofty's needs. Also no offense to `pastey`'s maintainer, but I'm much more trusting of an abandoned dtolnay crate anyway.

boul2gom:
I completely understand your perspective, and I fully agree that dtolnay's code is rock-solid. I'm not questioning the technical safety or the quality of the original paste crate.
My primary concern here isn't a functional bug, but the downstream impact on CI/CD pipelines. Because paste is now officially flagged with RUSTSEC-2024-0436 (Unmaintained), any project depending on lofty will start failing their automated security checks (like cargo-audit or Dependabot/Renovate).
While paste works fine, keeping it means every downstream user with strict CI requirements will be forced to manually add an exception to their security scanners just to use lofty.

In that case, what would be your official recommendation for downstream users dealing with these cargo-audit failures?

boul2gom:
For context on what other major crates are doing, several large projects have already made the switch to pastey pragmatically, specifically to avoid this downstream CI friction.
For instance, modelcontextprotocol/rust-sdk recently replaced paste with pastey as a drop-in replacement solely to resolve this RUSTSEC advisory.

I completely share your trust in dtolnay's code, but following the example of these larger ecosystem crates might save all our downstream users a lot of configuration headaches.

Serial-ATA:
> Because paste is now officially flagged with RUSTSEC-2024-0436 (Unmaintained)

Yeah I get where you're coming from, but I don't think that's reason enough to swap out `paste`. These RUSTSEC unmaintained advisories are really unfortunate, especially in this case where the crate isn't really "unmaintained", it's just ["finished"](https://github.com/rustsec/advisory-db/pull/2215#issuecomment-2709315704).

I'd really just urge people to ignore that advisory in their deny configs. Especially with it being one of the most downloaded crates, it's likely to show up in your dependency tree regardless.

PR #626, opened 2026-03-08
ItemKey: Add `ItemKey::MusicBrainzReleaseType`

matteron:
Hiya, needed this for reading album types (album vs ep vs single, etc).

https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html#id32

Not entirely sure if I did this correctly, I followed #573 as a template.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **ItemKey**: `ItemKey::MusicBrainzReleaseType` ([PR](https://github.com/Serial-ATA/lofty-rs/pull/626))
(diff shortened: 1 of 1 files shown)


Serial-ATA (changes requested):
LGTM, just one correction on the ID3v2 mapping

matteron:
I wasn't sure if this counted as a musicbrainz key or not.  For future reference what delineates that a field is exclusively a MusicBrainz field?

Serial-ATA:
AFAICT, there isn't really a "standard" field for release types outside of MusicBrainz. It looks like a lot of apps just make use of those MusicBrainz-specific fields (though I guess `RELEASETYPE` in Vorbis Comments could be generic). But now that you mention it, `ItemKey::MusicBrainzReleaseType` might be a better name just in case something else comes up.

matteron:
Let me know if you want the description comment changed, I wasn't sure how exactly to explain it without just saying "The release type is the album's type"

PR #644, opened 2026-04-14
Fix handling of undersized ID3 chunks in AIFF and WAV

acrilique:
Skip parsing of ID3 chunks with a size of zero to prevent unexpected EOF errors during header parsing.
Some aiff files sent by a friend (presumably generated by their DAW) caused this, so I wanted to get rid of the error.

This was AI-assisted!

Sample file: https://pixeldrain.net/u/qzEXaYGm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.24.0] - 2026-04-12
+
 ### Added
 
 - **APE** ([issue](https://github.com/Serial-ATA/lofty-rs/issues/631)) ([PR](https://github.com/Serial-ATA/lofty-rs/pull/633)):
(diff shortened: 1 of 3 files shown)


Serial-ATA (changes requested):
Good catch, thanks!

I think it'd be better to move this into `Chunk::id3_chunk()`:

https://github.com/Serial-ATA/lofty-rs/blob/2341386035be969a5590cd1c488bf91786fc365a/lofty/src/iff/chunk.rs#L271

It'd also be a good candidate for `FakeTag` errors in strict mode, something like:

```rust
if content.len() < 10 {
	log::warn!("ID3 chunk too small to contain a valid header");
	if parse_options.parsing_mode == ParsingMode::Strict {
		err!(FakeTag);
	}
	return Ok(None);
}
```

acrilique:
@Serial-ATA ptal

Serial-ATA:
:monocle_face: looks like I messed up that force push. I'll make a new PR

Serial-ATA:
#645

PR #649, opened 2026-04-21
misc: replace `OnceLock` with `LazyLock`

### LINKED ISSUE #441: Switch to `std::sync::LazyLock`
Usage of `once_cell::sync::Lazy` was replaced with `std::sync::OnceLock` awhile ago:

https://github.com/Serial-ATA/lofty-rs/blob/1505ddf69fd7df686bdf37e7306352011552c184/lofty/src/id3/v2/util/upgrade.rs#L41-L52

In 1.80.0, [`std::sync::LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) was finally stabilized. Now in that example, we can just go back to using `V2_KEYS` directly instead of having functions to access it.

This is blocked right now since 1.80.0 just came out and it really isn't that important to do anyway.

loxoron218:
I replaced `std::sync::OnceLock` with `std::sync::LazyLock` for static initializations in `upgrade.rs`, `write/mod.rs`, `resolve.rs`, and `tag/item.rs` to simplify lazy initialization logic and fix #441 

- `lofty/src/tag/items/popularimeter.rs` uses `OnceLock` with `.set()` and can't be converted to `LazyLock` without bigger API changes. That's why I left it :)

### DIFF
--- a/lofty/src/id3/v2/write/mod.rs
+++ b/lofty/src/id3/v2/write/mod.rs
@@ -20,21 +20,18 @@ use std::sync::LazyLock;
 use byteorder::{BigEndian, LittleEndian, WriteBytesExt};
 
 // In the very rare chance someone wants to write a CRC in their extended header
-fn crc_32_table() -> &'static [u32; 256] {
-	static INSTANCE: LazyLock<[u32; 256]> = LazyLock::new(|| {
-		let mut crc32_table = [0; 256];
+static CRC32_TABLE: LazyLock<[u32; 256]> = LazyLock::new(|| {
+	let mut crc32_table = [0; 256];
 
-		for n in 0..256 {
-			crc32_table[n as usize] = (0..8).fold(n as u32, |acc, _| match acc & 1 {
-				1 => 0xEDB8_8320 ^ (acc >> 1),
-				_ => acc >> 1,
-			});
-		}
+	for n in 0..256 {
+		crc32_table[n as usize] = (0..8).fold(n as u32, |acc, _| match acc & 1 {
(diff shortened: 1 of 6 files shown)


lofty/src/id3/v2/write/mod.rs
```diff
@@ -15,14 +15,13 @@ use crate::util::io::{FileLike, Length, Truncate};
 
 use std::io::{Cursor, Read, Seek, SeekFrom, Write};
 use std::ops::Not;
-use std::sync::OnceLock;
+use std::sync::LazyLock;
 
 use byteorder::{BigEndian, LittleEndian, WriteBytesExt};
 
 // In the very rare chance someone wants to write a CRC in their extended header
 fn crc_32_table() -> &'static [u32; 256] {
```
Serial-ATA:
```suggestion
static CRC32_TABLE: LazyLock<[u32; 256]> = LazyLock::new(|| {
```

I tend to prefer using statics directly rather than wrapping them in a function

lofty/src/resolve.rs
```diff
@@ -36,8 +36,8 @@ pub trait FileResolver: Send + Sync + AudioFile {
 type ResolverMap = HashMap<&'static str, &'static dyn ObjectSafeFileResolver>;
 
 pub(crate) fn custom_resolvers() -> &'static Arc<Mutex<ResolverMap>> {
```
Serial-ATA: Same here

Serial-ATA (changes requested):

Serial-ATA (approved):
Thanks!

PR #665, opened 2026-06-19
ItemKey: Add EBU-R128 mappings

ugniussle:
Opus defines the tags R128_TRACK_GAIN and R128_ALBUM_GAIN for usage with their header gain values ([RFC](https://datatracker.ietf.org/doc/html/rfc7845#section-5.2.1)).
I added the tags to the Vorbis comment tag map, seemed to work fine with my test file.


Serial-ATA:
Thanks!

PR #666, opened 2026-06-20
Opus: Add `OpusProperties::output_gain_db()`

ugniussle:
Save the Opus output gain in `OpusProperties` and add a function for returning that value converted to dB.


Serial-ATA:
Hey @ugniussle heads up on this and #665, it doesn't look like github is recognizing you as the author of these commits

ugniussle:
I fixed it.

Serial-ATA:
Thanks!

PR #672, opened 2026-07-08
docs: expand documentation for Tag utility methods, ItemKey, and Accessor

### LINKED ISSUE #357: Better document existing utility methods
Having looked through multiple projects over the years, it's clear that Lofty's API either isn't the best documented or the easiest to understand. Some patterns keep showing up, namely:

* Not using {[`Tag::take_strings`](https://docs.rs/lofty/latest/lofty/struct.Tag.html#method.take_strings), [`Tag::get_strings`](https://docs.rs/lofty/latest/lofty/struct.Tag.html#method.get_strings)}
	* Many projects could use `take_strings` instead of repeatedly cloning, since most don't write tags back to files.
	* Additionally, many projects could benefit from not having to iterate `TagItem`s and check the `ItemValue`. `get_strings` exists for that purpose.
	* Finally, I have seen some projects that don't know about [`ItemKey`](https://docs.rs/lofty/latest/lofty/enum.ItemKey.html). That makes it appear as though Lofty is only capable of providing the items accessible through [`Accessor`](https://docs.rs/lofty/latest/lofty/trait.Accessor.html). (!!!)
* Not using [`Tag::get_picture_type`](https://docs.rs/lofty/latest/lofty/struct.Tag.html#method.get_picture_type)
	* Many will iterate all pictures to find the one they want. That isn't necessary.

I consider this an issue with the docs, not with the projects. I believe there are other things I have found in the past, but this is what I could think of for now.

AZERDSQ131:
Addresses #357: several patterns keep showing up in downstream projects because the relevant docs are too terse to point people at the right API.

- `Accessor`: added a note clarifying it only covers a small common subset of fields, and pointing to `ItemKey`/`Tag::get_strings()`/`Tag::take_strings()`/`Tag::get_items()` for anything else — the trait's doc gave no hint that more exists.
- `ItemKey`: expanded the one-line doc to explain its role as the primary way to reach fields beyond `Accessor`, with an example, since (per the issue) some users don't realize `ItemKey` exists at all and assume `Accessor` is the full API surface.
- `Tag::get_strings()`: documented that it exists specifically to avoid manually iterating `get_items()` and matching on `ItemValue`, added an example, and cross-referenced `take_strings()`.
- `Tag::take_strings()`: documented when to prefer it over `get_strings()` (avoids cloning when the tag won't be written back), added an example.
- `Tag::get_picture_type()`: documented that it exists to avoid manually iterating `pictures()` and comparing `pic_type()`, added an example.

No behavior changes — doc comments and doctests only.

## Testing

- `cargo build -p lofty`
- `cargo test -p lofty --doc` (180 passed, 0 failed, including the new/updated examples)
- `RUSTDOCFLAGS="-D warnings" cargo doc -p lofty --no-deps` (clean, no broken/redundant intra-doc links)

Fixes #357

### DIFF
--- a/lofty/src/tag/accessor.rs
+++ b/lofty/src/tag/accessor.rs
@@ -35,6 +35,12 @@ macro_rules! accessor_trait {
 		/// Note that for tag formats supporting multiple values, the behavior of any setter methods is
 		/// to **overwrite**, not append. If multi-value support is needed, consider using the format-specific methods.
 		/// For example: [`Tag::push()`] or [`VorbisComments::push()`].
+		///
+		/// **`Accessor` is *not* an exhaustive list of everything a tag can store.** It only
+		/// covers a handful of fields that are common across formats. For anything else
+		/// (custom fields, format-specific fields, or fields that support multiple values),
+		/// use an [`ItemKey`](crate::tag::ItemKey) directly with methods such as
+		/// [`Tag::get_strings()`], [`Tag::take_strings()`], or [`Tag::get_items()`].
 		pub trait Accessor {
 			$(
 				accessor_trait! { @GETTER [$($name)+] $($ty),+ }
--- a/lofty/src/tag/item.rs
+++ b/lofty/src/tag/item.rs
@@ -482,6 +482,24 @@ macro_rules! gen_item_keys {
 		#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
 		#[non_exhaustive]
 		/// A generic representation of a tag's key
+		///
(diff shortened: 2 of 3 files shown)


Serial-ATA:
https://github.com/Serial-ATA/lofty-rs/blob/main/CONTRIBUTING.md#note-on-ai

AZERDSQ131:
Understood, and sorry for the noise — I hear the concern about fully AI-generated contributions with no significant human input, and I won't submit further PRs of that kind here. Thanks for the clear policy and for taking the time to point to it rather than just closing silently.

Issue #610, opened 2026-02-14
Document that `remove_<tag>` and `TagType::remove_from` are not equivalent


Serial-ATA:
`<File>::remove_<tag>` will only remove the tag in memory, it does **not** affect the final write. `TagType::remove_from()` will remove the tag on disk.

For example:

```rust
let mut file = File::open("foo.flac").unwrap();
let mut flac = FlacFile::read_from(&mut file).unwrap();
flac.remove_id3v2();
```

Will remove the tag from the `FlacFile`. That's useful if the tags from the `FlacFile` need to be copied to another file, without first needing to strip the ID3v2 tags from it.

```rust
TagType::Id3v2.remove_from_path("foo.flac").unwrap();

// or

let mut file = File::open("foo.flac").unwrap();
TagType::Id3v2.remove_from(&mut file).unwrap();
```

*This* is what actually removes the ID3v2 tag on disk.

samanshaiza004:
I'd like to work on #610 and #357. Since they're both documentation improvements around the metadata/tagging API, would you prefer one PR that addresses both, or separate PRs? I'm happy to do whichever is easier to review.

Serial-ATA:
Hey @samanshaiza004, thanks for picking these up! This one should be pretty small, so feel free to combine them into a single PR if you want. But since they're just doc revisions I'm fine reviewing either way :)

PR #677, opened 2026-07-14
docs: improve utility method docs, clarify remove_<tag> vs TagType::remove_from

### LINKED ISSUE #357: Better document existing utility methods

samanshaiza004:
closes #357 and #610 and did a tiny spelling pass (2 additional files affected)

The goal was to document the intended usage of the convenience APIs.

never written docs before so pls shoot down if not up to spec.

### DIFF
--- a/lofty/src/config/parse_options.rs
+++ b/lofty/src/config/parse_options.rs
@@ -201,7 +201,7 @@ pub enum ParsingMode {
 	///
 	/// This mode will attempt to fill in any holes where possible in otherwise valid, spec-compliant input.
 	///
-	/// NOTE: A readable input does *not* necessarily make it writeable.
+	/// NOTE: A readable input does *not* necessarily make it writable.
 	///
 	/// ## Examples of behavior
 	///
--- a/lofty/src/file/tagged_file.rs
+++ b/lofty/src/file/tagged_file.rs
@@ -273,6 +273,8 @@ pub trait TaggedFileExt {
 
 	/// Removes a specific [`TagType`] and returns it
 	///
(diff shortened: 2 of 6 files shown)


Serial-ATA:
Thanks! I made a few changes. Notably adding the docs to the generated `remove_<tag>` methods from `lofty_attr`

PR #682, opened 2026-07-21
ID3v2: Fix double unsynchronisation of frames when the tag header flag is set

### LINKED ISSUE #678: ID3v2.4: frames are de-unsynchronised twice when both the tag header and frame unsync flags are set, corrupting APIC pictures
### Reproducer

Generate a short silent MP3 next to the project:

```
ffmpeg -f lavfi -i anullsrc=r=44100:cl=mono -t 0.1 -c:a libmp3lame -b:a 64k -write_xing 0 -map_metadata -1 silence.mp3
```

Then run this against lofty 0.24.0. It builds a spec-conforming v2.4 tag with an unsynchronised APIC frame, the shape Bandcamp writes, and checks the picture bytes survive the roundtrip:

```rust
use lofty::prelude::*;

/// Stuff a zero after every 0xFF that precedes 0x00 or a sync-shaped byte,
/// the unsynchronisation scheme from id3v2.4.0-structure section 6.1.
fn unsynchronise(data: &[u8]) -> Vec<u8> {
    let mut out = Vec::new();
    for (i, b) in data.iter().enumerate() {
        out.push(*b);
        if *b == 0xFF && data.get(i + 1).is_some_and(|n| *n == 0x00 || *n >= 0xE0) {
            out.push(0x00);
        }
    }
    out
}

fn synchsafe(n: u32) -> [u8; 4] {
    [(n >> 21) as u8 & 0x7F, (n >> 14) as u8 & 0x7F, (n >> 7) as u8 & 0x7F, n as u8 & 0x7F]
}

fn main() {
    // Picture bytes containing an ff 00 pair, which real JPEGs are full of.
    let image = [0xFF, 0xD8, 0xFF, 0xE0, 0xFF, 0x00, 0x59, 0xFF, 0xD9];

    // APIC body: latin1 encoding, mime, front cover, empty description.
    let mut body = vec![0x00];
    body.extend(b"image/jpeg\0");
    body.push(3);
    body.push(0);
    body.extend(image);

    // Frame with unsynchronisation + data length indicator flags, the shape
    // Bandcamp's tagger writes.
    let stored = unsynchronise(&body);
    let mut frame = b"APIC".to_vec();
    frame.extend(synchsafe(stored.len() as u32 + 4));
    frame.extend([0x00, 0x03]);
    frame.extend(synchsafe(body.len() as u32));
    frame.extend(&stored);

    // v2.4 tag with the header unsynchronisation flag set, as the spec
    // requires when all frames carry the frame-level flag.
    let mut tag = b"ID3\x04\x00\x80".to_vec();
    tag.extend(synchsafe(frame.len() as u32));
    tag.extend(&frame);

    let mut mp3 = tag;
    mp3.extend(std::fs::read("silence.mp3").unwrap());

    let path = std::env::temp_dir().join("lofty-unsync-repro.mp3");
    std::fs::write(&path, &mp3).unwrap();

    let file = lofty::read_from_path(&path).unwrap();
    let picture = file.tags().iter().flat_map(|t| t.pictures()).next().unwrap();
    println!("expected: {:02x?}", image);
    println!("got:      {:02x?}", picture.data());
    assert_eq!(picture.data(), image, "picture data corrupted");
}
```

Output:

```
expected: [ff, d8, ff, e0, ff, 00, 59, ff, d9]
got:      [ff, d8, ff, e0, ff, 59, ff, d9]
thread 'main' panicked: picture data corrupted
```

The `ff 00 59` in the source became `ff 59`. On a real cover this is dozens of corruptions per image and the JPEG no longer decodes.

### Summary

When an ID3v2.4 tag sets the unsynchronisation flag in the tag header (0x80) and the frames also carry their own unsynchronisation flag (0x02), lofty applies the de-unsynchronisation twice: `parse_id3v2` wraps the whole tag in an `UnsynchronizedStream` ([`src/id3/v2/read.rs`](https://github.com/Serial-ATA/lofty-rs/blob/main/lofty/src/id3/v2/read.rs#L32-L36)), and the frame reader then wraps the frame content in a second `UnsynchronizedStream` for the frame flag. Each pass strips one `00` after every `ff`, so an encoded `ff 00 00` (which should decode to `ff 00`) collapses to `ff`.

For text frames this mostly goes unnoticed, but APIC content is binary and real JPEGs are full of `ff 00` pairs, so the returned picture data is corrupt and fails to decode. I found that Bandcamp's MP3 tagger writes exactly this (v2.4, tag flag 0x80, APIC frame flags 0x03). In my music library about 13% of MP3s are affected and every one of their embedded covers comes back mangled.

Per the v2.4 spec the scheme is applied per frame, and the tag header bit just indicates that all frames are unsynchronised, so the tag-level pass and the frame-level pass are the same operation done twice.

Reproduced on 0.24.0; the same two code paths are present on current main.

Two possible fixes, both resolve the double pass:

- Skip the frame-level `UnsynchronizedStream` when the tag-level stream is already active. This also keeps handling nonconforming v2.4 writers that unsynchronised the tag as one stream without setting frame flags.
- Treat the v2.4 tag header flag as informational (per spec it only summarises the frame flags) and only apply the scheme per frame. Cleaner against the spec, but drops support for the nonconforming whole-tag writers the current tag-level path happens to handle.

Right now in my own application I've separated out the cover loading by essentially going with the first option which is how I was able to confirm the issue.

I'd be more than happy to provide a real affected file if that helps (though it's music, so I don't want to share/attach it directly)!

### Expected behavior

Frame content is de-unsynchronised exactly once: an encoded `ff 00 00` decodes to `ff 00`, and the APIC picture bytes come back identical to what the encoder embedded, so covers decode cleanly.

### Assets

_No response_

attilagyorffy:
Fixes #678.

## The bug

When an ID3v2.4 tag sets the header unsynchronisation flag (`0x80`) and its frames also carry the per-frame unsynchronisation flag (`0x02`), lofty de-unsynchronises the frame content twice: once for the tag-level stream in `parse_id3v2`, and again for the frame flag in the frame reader. Each pass strips a `00` after every `0xFF`, so an encoded `FF 00 00` collapses to `FF`.

Text frames mostly survive this, but `APIC` pictures are binary and real JPEGs are full of `FF 00` pairs, so covers come back corrupt and no longer decode. This is the shape Bandcamp's MP3 tagger writes.

## The fix

Skip the frame-level unsynchronisation pass when the tag-level stream is already active, so frame content is decoded exactly once. The tag header's unsynchronisation flag is threaded down to the frame reader; the existing compression/encryption/plain arms then read the already-decoded bytes. Nonconforming writers that unsynchronise the whole tag without setting per-frame flags keep working, since that path is unchanged.

## Test

Added `unsynchronised_apic_is_decoded_once`, which builds the exact tag shape from #678 and checks the `APIC` bytes survive the round-trip. It fails on `main` (`FF 00 59` decodes to `FF 59`) and passes with the fix. The rest of the suite is unaffected.

Thanks to @zealsprince for the clear report and reproducer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - Fixed `Id3v2Tag::remove_disk_total()`, which incorrectly preserved the track number rather than disk number ([issue](https://github.com/Serial-ATA/lofty-rs/issues/656)) ([PR](https://github.com/Serial-ATA/lofty-rs/pull/657))
   - Fixed handling of encryption method symbols when writing ([issue](https://github.com/Serial-ATA/lofty-rs/issues/656)) ([PR](https://github.com/Serial-ATA/lofty-rs/pull/657))
   - Fixed parsing of extended headers in ID3v2.3 ([issue](https://github.com/Serial-ATA/lofty-rs/issues/656)) ([PR](https://github.com/Serial-ATA/lofty-rs/pull/657))
+  - Fixed double unsynchronisation of ID3v2.4 frames when the tag header unsynchronisation flag is set, which corrupted binary frames such as `APIC` ([issue](https://github.com/Serial-ATA/lofty-rs/issues/678)) ([PR](https://github.com/Serial-ATA/lofty-rs/pull/682))
 - **MusePack**: ID3v1 tags are now correctly treated as read-only
 - **MP4**: Fixed data loss in multi-value atom conversions ([issue](https://github.com/Serial-ATA/lofty-rs/issues/675)) ([PR](https://github.com/Serial-ATA/lofty-rs/pull/676))
 
--- a/lofty/src/id3/v2/frame/read.rs
+++ b/lofty/src/id3/v2/frame/read.rs
@@ -24,6 +24,7 @@ impl ParsedFrame<'static> {
 	pub(crate) fn read<R>(
 		reader: &mut R,
 		version: Id3v2Version,
+		tag_unsynchronisation: bool,
 		parse_options: ParseOptions,
 	) -> Result<Self, FrameParseError>
 	where
@@ -149,10 +150,17 @@ impl ParsedFrame<'static> {
 			// * unsynchronized + compressed
 			// * unsynchronized + encrypted
 			// * unsynchronized
+			//
+			// When the tag header sets the unsynchronisation flag, `parse_id3v2` already
+			// wraps the whole tag in an `UnsynchronizedStream`, so the frame content
+			// reaching us here is de-unsynchronised. Applying the frame-level scheme again
+			// would strip a second `00` after every `0xFF` and corrupt binary frames such
+			// as APIC. Skip the frame-level pass in that case and let the remaining arms
+			// read the already-decoded bytes (see issue #678).
 			FrameFlags {
 				unsynchronisation: true,
 				..
-			} => {
+			} if !tag_unsynchronisation => {
 				let mut unsynchronized_reader = UnsynchronizedStream::new(reader);
 
 				if flags.compression {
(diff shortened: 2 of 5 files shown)


Serial-ATA:
superseded by #690

PR #689, opened 2026-08-05
examples: Collapse nested `if`s in `tag_stripper`

jfietz:
Following the `cargo fmt`/`cargo clippy`/`cargo doc` step in
CONTRIBUTING.md before opening a PR, the command`cargo clippy --all-targets` fails on current stable (clippy 0.1.97) with two issues:

```
clippy::collapsible_if` errors in `examples/tag_stripper.rs`. Since let-chains
are available on the current MSRV (1.89, edition 2024), the nested conditions
collapse into a single `if`.
```

CI does not currently catch this, as it runs `cargo clippy --all-features`
without `--all-targets`, so the examples are never linted. 

No behavioral change - a failure at any step still falls through to the
"Unexpected input" branch.

### DIFF
--- a/examples/tag_stripper.rs
+++ b/examples/tag_stripper.rs
@@ -41,14 +41,13 @@ fn main() {
 		print!("\nNumber to remove: ");
 		std::io::stdout().flush().unwrap();
 
-		if std::io::stdin().read_line(&mut input).is_ok() {
-			if let Ok(num) = str::parse::<usize>(input.trim()) {
-				if num < available_tag_types.len() {
-					to_remove = Some(num);
-					println!();
-					break;
-				}
-			}
+		if std::io::stdin().read_line(&mut input).is_ok()
+			&& let Ok(num) = str::parse::<usize>(input.trim())
+			&& num < available_tag_types.len()
+		{
+			to_remove = Some(num);
+			println!();
+			break;
 		}
 
 		input.clear();


Serial-ATA:
A single Clippy warning fix doesn't warrant a Claude co-author.

jfietz:
@Serial-ATA: well, it was split out of the big fix you closed without comment. Of a bug that corrupted  10% of my media library. I understand your stance, but both issues are real, both issues are fixed afterwards, I've iterated on the code as well.

If your stance is "no AI-edited contributed code" instead of "fine if clearly marked", please clarify that. I did spent not insignificant time and brain power on this, even if I used Claude as well

Serial-ATA:
I apologize that my AI stance isn't clear.

Please do recognize that this is a new thing for me. This year's issues and PRs have been almost entirely AI generated, so I tried adding a [small note](https://github.com/Serial-ATA/lofty-rs/blob/main/CONTRIBUTING.md#note-on-ai) to the CONTRIBUTING.md to maybe inspire people to be a little more human. That didn't stop anything.

Here's my POV: At this point I find these fully AI contributions to be flat out disrespectful. More often than not they end up being wrong, but of course the burden of determining that is thrown onto me since no one has the decency to read their code these days. This isn't calling _you_ out specifically. Maybe you'd be different and actually read my review comments before passing them off to Claude. But I don't have the time nor patience to risk going back and forth with an LLM again.

If you solved a problem by throwing a prompt at Claude, don't assume I couldn't just as easily do the same.

jfietz:
I am very sympathetic, and tried to be respectful by adding the clear disclaimers. As you say, sucks that that has helped absolutely nothing. 

The fix here is a simple contraction of the three nested ifs that was flagged when first checking out the repository and following the instructions, but nothing that would keep a contributor from helping.

My other PR is also a very simple single-digit line fix and it's the testing that is excessively verbose.

Thanks for doing your work in the open, may better times come soon.

Issue #445, opened 2024-08-25
Make use of padding when writing


Serial-ATA:
Right now we just retain existing padding and add some if it's missing.

- [ ] MP4: Utilize `free` atoms surrounding `ilst` atoms
  - [x] Calculate space available before writing
	  - When writing, all available space is taken into consideration. This includes **all** `free` atoms before AND after the `ilst`, as well as the `ilst` itself. Any remaining space is converted back to padding.
	    - This means if you save a 500 byte tag to the following file:
	      - `free` (250 bytes)
	      - `ilst` (250 bytes)
	      - `free` (250 bytes)
	    - The result will be:
	      - `ilst` (500 bytes)
	      - `free` (250 bytes)
  - [ ] Do not rewrite the entire file
    - We should be able to update the tag as above *and* its offsets without having to rewrite the entire file.
- [ ] FLAC: Utilize `PADDING` blocks
  - We should be able to handle the cases:
    - [ ] No padding, write as normal and append to the end
    - [ ] Padding exists at the end, do not rewrite the file if it's enough
      - Still rewrite if the remaining padding is less than the `preferred_padding`
    - [ ] Padding exists between blocks, such as after `VORBIS_COMMENTS` and before a `PICTURE`
      - We should be able to grow the `VORBIS_COMMENTS` and leave `PICTURE` alone if possible. (?)
- [ ] ID3v2: Simply overwrite tags
  - ID3v2 supports padding, so if `tag_to_write.len() <= existing_tag.len()`, we can simply write `tag_to_write` at the beginning of the file and replace any remaining size with padding.
  - This *only* applies to formats where ID3v2 can be written to the beginning of the file (not IFF formats)

PR #695, opened 2026-08-12
FLAC: Avoid buffering the entire file during tag writes

### LINKED ISSUE #694: FLAC: Tag writes buffer the entire file in memory
### Reproducer

Write or modify a Vorbis comment on a large FLAC using Lofty's normal public save API and measure the process's peak resident memory.

On Lofty 0.25.0, the FLAC writer currently reads the complete file into memory before replacing the metadata range.

I reproduced this with a generated 357,463,057-byte (~341 MiB) FLAC.

Peak RSS while performing the metadata write:

- Lofty 0.25.0: ~532,952 KiB
- bounded-I/O prototype: ~8,240 KiB

Both outputs passed `flac -t` and preserved the FLAC audio MD5/signature.

### Summary

Writing metadata to a FLAC currently makes memory usage scale with the size of the entire audio file rather than the size of the metadata.

The current FLAC writer reads the whole file into a `Vec`, modifies the metadata region, truncates the original file, and writes the full buffer back.

This is noticeable with large FLAC files. A ~341 MiB test file caused roughly 533 MiB peak RSS for a simple metadata write.

I prototyped an alternative that keeps Lofty's existing FLAC parsing and metadata encoding, but replaces the final whole-file buffer/splice with an in-place range replacement using a fixed 64 KiB movement buffer.

On the same file that reduced peak RSS to about 8 MiB while preserving audio integrity.

This seems related to #445, although the narrower issue here is whole-file memory buffering rather than full FLAC padding optimization.

### Expected behavior

FLAC metadata writes should not require buffering the complete encoded audio payload in memory.

Memory usage should primarily depend on the metadata being parsed/encoded plus a bounded I/O buffer.

### Assets

AI disclosure: I used AI assistance during the investigation, prototyping, test design, and drafting of this report. The measurements and validation above were reproduced locally. If this proceeds to a PR, I intend to review and understand the patch myself and participate directly in the review rather than treating the AI output as authoritative. I'm sorry if this is annoying, tbh i'll have to patch this in my code everytime you guys update, otherwise i probably wouldn't bother y'all. cheers.

No copyrighted media is required to reproduce this.

The test file was generated from random PCM data.

I also have a tested one-file prototype patch against Lofty 0.25.0 that:

- preserves the existing public API;
- adds no dependencies;
- adds no unsafe code;
- keeps Lofty's existing FLAC metadata parsing/encoding;
- moves the remaining file contents in bounded 64 KiB chunks;
- passes Rust 1.89 MSRV and the current test, Clippy, fmt and docs CI commands;
- passes real FLAC grow/shrink tests with preserved audio MD5/signature.

I can submit the patch as a PR if this approach is acceptable.

sxntaxis:
closes #694

This keeps the existing FLAC parsing and metadata encoding, but replaces the final whole-file buffer/splice with a bounded in-place range replacement.

- Equal-sized metadata is overwritten in place.
- When metadata grows, the file is extended and the remaining contents are moved backwards in 64 KiB chunks before writing the replacement.
- When metadata shrinks, the remaining contents are moved forwards and the file is truncated afterward.

The FLAC resize regression now covers FLAC directly. I also checked the historical #549/#607/#608/#640 failure modes, including no-padding/minimal layouts, last-metadata-block handling, leading ID3 data, and metadata shrink/grow. The large-file reproducer from #694 still passes `flac -t` with the audio signature preserved while avoiding whole-file memory use.

This does not try to solve #445: existing FLAC padding is still handled the same way, so audio may still need to move when metadata size changes.

AI disclosure: I used AI assistance during investigation, prototyping, test design, and drafting. I reviewed the behavior and failure modes, reproduced the tests and measurements locally, and will participate directly in review.

### DIFF
--- a/lofty/src/flac/write.rs
+++ b/lofty/src/flac/write.rs
@@ -145,40 +145,27 @@ where
 	if will_write_padding && let Some(preferred_padding) = write_options.preferred_padding {
 		log::warn!("File is missing a PADDING block. Adding one");
 
+		let metadata_len = blocks
+			.iter()
+			.map(|block| u64::from(block.len()))
+			.sum::<u64>();
+		let old_metadata_len = metadata_range.end - metadata_range.start;
+		let available_padding = old_metadata_len
+			.saturating_sub(metadata_len)
+			.saturating_sub(Block::BLOCK_HEADER_SIZE as u64);
+		let padding_len = available_padding
+			.max(u64::from(preferred_padding))
+			.min(u64::from(Block::MAX_CONTENT_SIZE));
+		let padding_len = usize::try_from(padding_len).map_err(|_| SizeMismatchError)?;
+
 		// `PADDING` always goes last
-		let mut padding_block = Block::new_padding(preferred_padding as usize)?;
-		padding_block.last = true;
-
-		blocks.push(padding_block);
+		blocks.push(Block::new_padding(padding_len)?);
 	}
 
-	let mut encoded_metadata = encode_blocks(&blocks)?;
-
-	if will_write_padding
-		&& let Some(preferred_padding) = write_options.preferred_padding
-		&& (encoded_metadata.len() as u64) < metadata_range.end - metadata_range.start
-	{
-		// Keep the metadata region with the PADDING block already created above. Its header is
-		// part of the retained span, so only the remaining bytes become block content.
-		let padding_index = blocks.len() - 1;
-		let non_padding_len = encoded_metadata.len() - blocks[padding_index].len() as usize;
-		let available = metadata_range.end - metadata_range.start - non_padding_len as u64;
-		let padding_span = available.max(u64::from(preferred_padding) + 4).max(4);
-		let padding_content = usize::try_from(padding_span - 4).map_err(|_| SizeMismatchError)?;
-		let mut padding_block = Block::new_padding(padding_content)?;
-		padding_block.last = true;
-		blocks[padding_index] = padding_block;
-
-		encoded_metadata = encode_blocks(&blocks)?;
-	} else if let Some(block) = blocks.last_mut() {
+	if let Some(block) = blocks.last_mut() {
 		block.last = true;
-		encoded_metadata = encode_blocks(&blocks)?;
 	}
 
-	Ok(replace_range(&mut file, metadata_range, &encoded_metadata)?)
-}
-
-fn encode_blocks(blocks: &[Block]) -> Result<Vec<u8>, FileEncodingError> {
 	let mut encoded_metadata = Vec::new();
 	for block in blocks {
 		block.write_to(&mut encoded_metadata)?;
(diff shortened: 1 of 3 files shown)


lofty/src/flac/write.rs
```diff
@@ -169,11 +166,130 @@ where
...
+			// The ranges overlap, so move the tail backwards from EOF before writing metadata.
+			extend_storage(file, difference, &buffer)?;
+			shift_right(file, range.end, file_len, difference, &mut buffer)?;
+		},
+		std::cmp::Ordering::Less => {
+			let difference = old_len - replacement_len;
+			// Move forwards from the metadata boundary so writes cannot clobber unread tail data.
+			shift_left(file, range.end, file_len, difference, &mut buffer)?;
+			file.truncate(file_len - difference)?;
+		},
```
Serial-ATA, line 223: I think we should only be doing a left shift if `preferred_padding` is `None`. Otherwise, we should always fill any gap with the `PADDING` block created above. Since `preferred_padding` is just a hint, if the gap is:

* Less than `preferred_padding` - round up to `preferred_padding`
* Greater than `preferred_padding` - fill the remaining space with padding

Serial-ATA (changes requested):

lofty/src/flac/write.rs
sxntaxis: Addressed in [944ca9d5](https://github.com/Serial-ATA/lofty-rs/pull/695/changes/944ca9d5afbcf048a87493ba909a6568b21067f4). Shrinking now only shifts the tail when preferred padding is disabled; with preferred padding, the existing padding block keeps the available gap or grows to the requested minimum. Added regressions for both cases and for the no-padding shrink path.

lofty/src/flac/write.rs
```diff
@@ -155,10 +152,33 @@ where
 		blocks.push(padding_block);
 	}
 
-	if let Some(block) = blocks.last_mut() {
-		block.last = true
+	let mut encoded_metadata = encode_blocks(&blocks)?;
```
Serial-ATA: You're encoding the blocks twice. You can use `Block::len()` to get the cumulative length of the new blocks without encoding.

lofty/src/flac/write.rs
```diff
@@ -155,10 +152,33 @@ where
...
+
+	if will_write_padding
+		&& let Some(preferred_padding) = write_options.preferred_padding
+		&& (encoded_metadata.len() as u64) < metadata_range.end - metadata_range.start
```
Serial-ATA: Padding shouldn't have any length preconditions. If `encoded_metadata.len()` is greater than the current gap, we still need to write `preferred_padding` as long as `will_write_padding` is `true`.

Serial-ATA (changes requested):

Serial-ATA (approved):

Issue #711, opened 2026-09-06
Allow passing of WriteOptions.parse_options to VerifiedFile::new() and their use. (similar to issue #668 (1))


mitchtaylormx:
### Reproducer

(Summary)
Lofty's VerifiedFile probe doesn't receive the caller's WriteOptions.parse_options, so it silently reverted to the 1,024-byte default set in ParseOptions::DEFAULT_MAX_JUNK_BYTES. The passing of WriteOptions.parse_options to VerifiedFile::new() would allow flexibility of file compatibility. Same as the #1 in [https://github.com/Serial-ATA/lofty-rs/issues/668] which was closed, but only the second issue might have been done?

(Background)
Some of my MP3 files had more than 1024 bytes of "Junk Data" after the tag (v2.3) and before the audio section.  I wanted to figure out why they were failing when the tag was being updated so I started looking into it and found that a manual edit with other software worked fine, but Lofty would just fail. Since Lofty uses only the default of 1024 bytes to search after the tag for the audio data the write was failing. If the ParseOptions::DEFAULT_MAX_JUNK_BYTES was changed to a higher number (4096?) or the WriteOptions.parse_options was used with a higher number then they would have written without error. 

(Reproducer)
Attempt to add a tag to an MP3 file with more than 1024 junk bytes after the tag and before the audio stream.

Code from my update to bliss-analyser (for Lofty 0.25.1):

```
   fn set_mp3_analysis_tag(
        path: &Path,
        analysis_tag: &str,
        value: &str,
    ) -> Result<(), Box<dyn std::error::Error>> {
        use lofty::id3::v2::{
            ExtendedTextFrame,
            Frame,
            Id3v2Tag,
        };

        use lofty::TextEncoding;
        use lofty::mpeg::MpegFile;

        let mut input = std::fs::File::open(path)?;

        let mut file = MpegFile::read_from(
            &mut input,
            ParseOptions::new(),
        )?;
        let tag = match file.id3v2_mut() {
            Some(tag) => tag,

            None => {
                file.set_id3v2(Id3v2Tag::new());
                file.id3v2_mut().unwrap()
            }
        };
        tag.insert(Frame::UserText(
            ExtendedTextFrame::new(
                TextEncoding::UTF8,
                Cow::Owned(analysis_tag.to_owned()),
                Cow::Owned(value.to_owned()),
            ),
        ));

        file.save_to_path(
            path,
            WriteOptions::default(),
        )?;
```

[...]

PR #713, opened 2026-09-07
Tests: Add `WriteOptions::parse_options` regression test

mitchtaylormx:
## Summary

Adds a regression test for issue #711 / PR #712.

The test exercises writing a tag to an MP3 containing more than the default 1024 bytes of leading junk, while supplying ParseOptions with max_junk_bytes(3000) through WriteOptions.

This test fails with the pre-#712 implementation because the internal initial probe ignored WriteOptions.parse_options, and passes with the fix from PR #712.

## Testing

* cargo test -p lofty --test files mpeg::test_tag_save_with_parse_options
* 1 passed

* cargo test -p lofty --test files
* 123 passed
* 0 failed
* 0 ignored


Serial-ATA (changes requested):
`tests/taglib` isn't the right place to put this, that dir only mirrors the tests from [TagLib](https://github.com/taglib/taglib). `tests/files/mpeg.rs` would be better

mitchtaylormx:
Got it moved. Sorry, thought about the other files tests before I put it in the wrong place...  It just seemed so similar to the taglib one I had placed it after not realizing that those are a standard test lib (or something like that...) Hope this helps. Regards and thanks for the quick updates. I'm looking forward to 0.25.2 to start integrating it, but I am also trying to run some speed tests after making a move from lofty 0.22.x...

Serial-ATA:
Thanks, LGTM after the formatting is fixed

> I am also trying to run some speed tests after making a move from lofty 0.22.x...

Have you noticed worse performance in newer versions?

mitchtaylormx:
My 'sed' command to move it wasn't perfect... I should have looked at it in the file... All better now.

The changes I made to move from 0.22.2 to 0.25.1 by replacing all the ::Unknown usage with file specific functions with the correct (I think) methods to, in particular, add a new non-standard tag seem a fair amount slower. It might be that I am reading multiple times (or some bad coding... ) when I should not be, or something else. Reads seem fast, the slowest seems to be writing to a flac with a new tag inserted.  I have some timing checks in the code I am still testing to see if I can find the slowdown...

Serial-ATA (approved):

PR #725, opened 2026-09-19
ID3v2: Correct growing WAV and AIFF stream sizes

### LINKED ISSUE #724: ID3v2: Growing a trailing WAV or AIFF chunk decreases the container size
## Reproducer

Start with a RIFF/WAVE or FORM/AIFF stream whose final chunk is an ID3v2 tag, then save a larger ID3v2 tag to the same stream. After the write, compare the 32-bit container size at bytes 4 through 7 with the actual stream length minus eight.

This affects Lofty 0.25.2, 0.25.3, and the current `main` branch.

## Summary

The trailing-chunk fast path subtracts tag growth from `updated_stream_len`:

```rust
updated_stream_len -= tag_chunk_size - existing_tag_len;
```

When the replacement is larger, the outer RIFF/FORM size becomes smaller than the file. A sufficiently large increase also underflows and panics in debug builds after the new tag bytes have been written.

Release Maestro found this while testing repeated WAV and AIFF metadata growth: https://github.com/floydnant/release-maestro/pull/166

## Expected behavior

Growing a trailing ID3v2 chunk should increase the outer RIFF/FORM stream length. The declared container size should equal the actual stream length minus eight after the write.

## Assets

No external asset is required. The bug reproduces with a minimal in-memory RIFF or FORM header and two successive ID3v2 writes.

floydnant:
Fix the trailing ID3 chunk fast path so tag growth increases the outer RIFF/FORM stream size.

Add regression coverage for WAV and AIFF using existing fixtures.

Closes #724.


Serial-ATA:
:facepalm: thanks, I thought there was a test for this already but looks like I only have MP3 covered: https://github.com/Serial-ATA/lofty-rs/blob/423ff714148726f1568e1ef127f525269c0e50a2/lofty/tests/files/resize.rs#L96-L98

When I get a chance I'll update those tests to get them running for all supported formats, then we can drop the tests from this PR

Serial-ATA:
Actually, I'll keep these tests around for now. Looks like FFmpeg (which is used in the resize tests) doesn't care about the stream length, so the broken containers would pass anyway. Might look into stricter validation tools in the future

